PR #7, opened 2014-07-15
Adding start of a basic pure cpython module build with CMake.

blowekamp:
I have added a directory with a basic CPython module which can be build with CMake. There is still work to do to get it working with PyCMake.setup. Additionally, it's not currently being tested. It should provide a simple example to experiment with how things should be done.

### DIFF
--- a/tests/purecpython_test1/CMakeLists.txt
+++ b/tests/purecpython_test1/CMakeLists.txt
@@ -1,0 +1,16 @@
+cmake_minimum_required (VERSION 2.8.1 FATAL_ERROR)
+
+project(PyCMakePureCPythonTest)
+
+#
+# Use standard CMake package to find Python. This should be passed
+# from PyCMake.
+#
+find_package(PythonInterp REQUIRED)
+find_package(PythonLibs REQUIRED)
+
+
+add_library(test1 MODULE src/test1.c)
+
+include_directories(${PYTHON_INCLUDE_DIRS})
+target_link_libraries(test1 ${PYTHON_LIBRARIES})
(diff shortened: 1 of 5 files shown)


opadron:
Hi, all.  So, it looks like I've been assigned to see this PR through to resolution, so I'm necro-ing this thread.

As far as basic CPython modules go, we have [jcfr/python-cmake-custom-extension](https://github.com/jcfr/python-cmake-custom-extension), which has the latest changes necessary to work with #30.

So, is there any further action that is necessary for this PR?

blowekamp:
Please take anything needed from here to help develop a comprehensive frame work for testing packaging and building with scikit build.

opadron:
We'll be adding example apps to the testing soon, so, let's keep this open in case we need to borrow anything here.

And then once we have a few apps in place and tested, we can close this up.

How does that sound?

jcfr:
> How does that sound?

Makes sense.

I also like the idea of having simple examples, adding directories like  `tests/samples/intocpython` could make sense.

opadron:
We are almost done working on a large change set in #43 that should add several test projects.  In particular, please have a look at [hello](https://github.com/scikit-build/scikit-build/pull/43/commits/abc9221db54243ffbba344e08ee6bcf407fbb175) and [hello-cython](https://github.com/scikit-build/scikit-build/pull/43/commits/6d99ec0268f688f6c059dc6db6cbb55c3c9ecfa7).

I suggest this issue be resolved and closed once #43 is merged.

What do you think, @blowekamp?

opadron:
I'm going to go ahead and close this.  Feel free to reopen if necessary.

PR #30, opened 2016-04-12
try a package-centric approach

opadron:
The idea is for package writers to develop their packages as if they were pure
  python, except with the addition of C/C++ source files scattered where
  desired.  The mapping of source files to libraries are provided by
  CMakeList.txt files, which must install the libraries within the source
  directory; at the locations where the imports in the pure python portions of
  the package expect them.

  This example follows popular practice for packages featuring a mix of pure and
  compiled code.  Most compiled functionality are isolated in self-contained
  modules that are built into shared libraries with names starting with an
  underscore (_).  These libraries are paired with a pure python module of the
  same name sans underscore.  In this module the compiled functionality is
  exposed with the desired wrappings, or more commonly, a direct pass-through
  import (from ._compiled import *).

  - Also improves cmake integration by directly passing PYTHON_INCLUDE_DIR and
    PYTHON_LIBRARY variables.
    - prioritizes dynamic over static libraries

### DIFF
--- a/skbuild/cmaker.py
+++ b/skbuild/cmaker.py
@@ -1,8 +1,11 @@
+import argparse
+import itertools
 import os
-import sys
+import os.path
 import platform
 import subprocess
-import argparse
+import sys
+import sysconfig
 
 from .platform_specifics import get_platform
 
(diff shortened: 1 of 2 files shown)


opadron:
See https://github.com/jcfr/python-cmake-custom-extension/pull/1 for a version of @jcfr's example application that should work with these changes.

blowekamp:
Sounds reasonable.

Please keep in mind that on OS X and Linux (elf binaries atleast), you are not suppose to link against the libpython library. The loaded module is suppose to get the symbols from the python executable. [1]

[1] https://github.com/SimpleITK/SimpleITK/blob/master/CMake/sitkTargetLinkLibrariesWithDynamicLookup.cmake

[2] https://mail.python.org/pipermail/distutils-sig/2016-February/028275.html

opadron:
Right, but cmake still looks for it, so it should be fine to at least point out which one it should look at.

For projects using skbuild - it'd be nice if the same behavior was present for the dll loader in windows, but I suspect not, leaving package devs in the scenario of keeping the library link line in their CMakeLists files and having to guard every one of them with `IF WINDOWS`.  Maybe there's some way to add a simple cmake function to hide that?

thewtex:
This PR LGTM.

Regarding the proposal: it sounds good, but the CMake configuration for new package authors should not be a focus, yet. A higher priority is to allow users to build projects that already have a CMake configuration, like SimpleITK, to be built through pip from source and to generate binary packages.

opadron:
@thewtex How much change should we be asking of existing packages in order to adopt skbuild?  Surely, little to no change is not realistic, and as long as we're asking that existing packages follow _some_ path to adoption, the distinction between new and existing packages isn't relevant.

If the intention, instead, is to somehow make SimpleITK pip installable just by slapping on a setup.py script and declaring skbuild as a dependency, then this PR still needs more work, imo.  In its current form, this PR requires packages to build and copy extension libraries to particular places, so that the manifest-parsing logic would create the right `package_data` spec for the setup call.  If we're not even going to ask existing packages for that much, then we'd need a way to "magically discover" what binaries a package's CMake build creates that's more reliable than going through a manifest and guessing.

jcfr:
> SimpleITK pip installable just by slapping on a setup.py

Yes

> discover" what binaries a package's CMake build creates 

The idea here would be to leverage components associated with install rules:

* By default all files associated with `RuntimeLibraries` (and `Runtime` ?) would be bundled in the wheel

* If needed, we could come up with a dedicated component name `PythonModules` ? 

If no components are associated with install rules, we could bundle everything in the wheel.

opadron:
I've had a chat with @jcfr, and we are in agreement that this PR takes scikit-build in the right direction.  We've also identified a few parts that need fixing.  I will update and ping everyone for a final review.

opadron:
@scikit-build/developers PTAL

jcfr:
LGTM. Please integrate.

jcfr:
LGTM. Can I integrate  ?

PR #31, opened 2016-04-12
CI integration

### LINKED ISSUE #24: Setup travis, appveyor and circleci

opadron:
fixes #24 (eventually...)


opadron:
Holy cow, that took way more work than it should have.
Let's hope the AppVeyor build just works. 🎱

opadron:
I knew it was too good to be true. :^)

opadron:
"No module named nose"... **progress!** 👍

opadron:
@scikit-build/developers I'm not sure what more can be done with AppVeyor.  The test builds run all the way up to the last line (`%PYTHON%\python.exe -m nose -w tests`) and then hang until they are killed after an hour.  I've tried everything I could think of to get this working, and googling has turned up very little.  I'm at my wit's end with this one.

jcfr:
Thanks for the update.

This most likely happen because of the python network tests requiring the user to explicitly authorize them by clicking "Allow" in a popup.

You could either exclude these tests or relax the firewall rules. 

Hth
Jc

opadron:
@jcfr Are we actually running test cases that require inbound connections from outside?  That's the only scenario where the firewall should even matter.

Installing cmake from chocolatey and the python packages via pip work just fine, so we know we can make connections to the web.

Am I missing something?

jcfr:
@opadron Some of the tests use the loopback and popup like [this one](http://maximumpcguides.com/windows-7/wp-content/uploads/2009/07/allow-program-windows-7-firewall-7-468x334.PNG) are probably shown.

opadron:
I figured out how to RDP to a hanging build worker.  This is what I found:
https://i.imgur.com/J5PKcPM.png

Looks like something's not right with compilers?

jcfr:
Looks like it is build with mingw and the library fail to load at runtime ?
If you look at GitHub.com/python-cmake-buildsystem, you could see how we do
the dashboard submission by part.

opadron:
Hooray! Some of the builds are successful!

Still looks like we have some problem with MinGW.  @jcfr I looked around cmake-python, but I'm not sure where to look for info about mingw.

jcfr:
Congratsss.

It was more about the dashboard script s in the dashboard branch.

opadron:
Assuming this last check passes, we'll have the following tests in place for Windows:

|VStudio|MinGW|Default Generator|PY2|PY3|32-bit|64-bit|Fortran|
|---|---|---|---|---|---|---|---|
|:white_check_mark:|||:white_check_mark:||:white_check_mark:||:x:|
|:white_check_mark:|||:white_check_mark:|||:white_check_mark:|:x:|
|:white_check_mark:||||:white_check_mark:|:white_check_mark:||:x:|
|:white_check_mark:||||:white_check_mark:||:white_check_mark:|:x:|
||:white_check_mark:||:white_check_mark:|||:white_check_mark:|:white_check_mark:|
||:white_check_mark:|||:white_check_mark:||:white_check_mark:|:white_check_mark:|
|||:white_check_mark:|:white_check_mark:|||:white_check_mark:|:x:|
|||:white_check_mark:||:white_check_mark:||:white_check_mark:|:x:|

### Reading the table
Each row is a build.
:white_check_mark: - Tested feature
:x: - Missing functionality (Microsoft does not offer a Fortran compiler).

opadron:
Curses!  We're so close!

opadron:
Ok, I'm having a look through an appveyor build-worker and also the appveyor documentation...and it looks like they don't have any fortran compilers at all.

It seems like we should just leave out the fortran tests entirely for the windows testing.

Thoughts?

opadron:
I've gone back to removing all Fortran testing from Appveyor.

That should put us at the following testing coverage on Windows:

|VStudio|MinGW|Default Generator|PY2|PY3|32-bit|64-bit|Fortran|
|---|---|---|---|---|---|---|---|
|:white_check_mark:|||:white_check_mark:||:white_check_mark:||:x:|
|:white_check_mark:|||:white_check_mark:|||:white_check_mark:|:x:|
|:white_check_mark:||||:white_check_mark:|:white_check_mark:||:x:|
|:white_check_mark:||||:white_check_mark:||:white_check_mark:|:x:|
||:white_check_mark:||:white_check_mark:|||:white_check_mark:|:x:|
||:white_check_mark:|||:white_check_mark:||:white_check_mark:|:x:|
|||:white_check_mark:|:white_check_mark:|||:white_check_mark:|:x:|
|||:white_check_mark:||:white_check_mark:||:white_check_mark:|:x:|

### Reading the table
Each row is a build.
:white_check_mark: - Tested feature
:x: - Missing functionality

opadron:
With CircleCI testing up and running, I am finally calling this PR ready for review.

@scikit-build/developers PTAL.

appveyor.yml
```diff
@@ -0,0 +1,170 @@
...
+  - ps: "ls \"C:/\""
+  - cmd: ECHO "Installed SDKs:"
+  - ps: "ls \"C:/Program Files/Microsoft SDKs/Windows\""
+
+  # Prepend newly installed Python to the PATH of this build (this cannot be
+  # done from inside the powershell script as it would require to restart
+  # the parent CMD process).
+  - cmd: "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
+
+  # Check that we have the expected version and architecture for Python
```
jcfr: It looks like the comment should be changed to
```
Display version and architecture for Python. That is useful when visually inspecting the log to identify potential issues.
```

appveyor.yml
```diff
@@ -0,0 +1,170 @@
...
+  # target Python version and architecture
+  - cmd: "%CMD_IN_ENV% pip install -r requirements.txt"
+  - cmd: "%CMD_IN_ENV% pip install -r requirements-dev.txt"
+
+build_script:
+  # Build the compiled extension
+  - cmd: "%CMD_IN_ENV% python setup.py build"
+
+test_script:
+  # workaround for CMake not wanting sh.exe on PATH for MinGW
```
jcfr: Is there already an issue in CMake tracker ?

appveyor.yml
```diff
@@ -0,0 +1,170 @@
...
+  # Archive the generated packages in the ci.appveyor.com build report.
+  - path: dist\*
+
+matrix:
+  fast_finish: true
+
+notifications:
+  - provider: GitHubPullRequest
+    auth_token:
+      secure: iXYaAwXJcHdnRGTeb1Gk1PembhlZleCGDK2oJgsiFvA=
```
jcfr: Is it the right location for this token ? Should this be added in the settings associated with the project ?

ci/appveyor/run-with-visual-studio.cmd
```diff
@@ -0,0 +1,88 @@
...
+:: MS Windows SDK for Windows 7 and .NET Framework 3.5 (SDK v7.0)
+::
+:: 32 bit builds, and 64-bit builds for 3.5 and beyond, do not require specific
+:: environment configurations.
+::
+:: Note: this script needs to be run with the /E:ON and /V:ON flags for the
+:: cmd interpreter, at least for (SDK v7.0)
+::
+:: More details at:
+:: https://github.com/cython/cython/wiki/64BitCythonExtensionsOnWindows
```
jcfr: Looks like the page has been renamed to https://github.com/cython/cython/wiki/CythonExtensionsOnWindows

ci/appveyor/run-with-visual-studio.cmd
```diff
@@ -0,0 +1,88 @@
+:: To build extensions for 64 bit Python 3, we need to configure environment
```
jcfr: Was this final originally copied from https://github.com/cython/cython/blob/master/appveyor/run_with_env.cmd ? 

May be the commit message or the first line of the file could mention it.

jcfr:
Very Nice. Great to see the CI returns green :smile:

appveyor.yml
jcfr: Thanks for your offline explanation. It is encrypted using `Account -> Encrypt data` available from Appveyor website after signing in.

appveyor.yml
opadron: I found this: https://cmake.org/Bug/view.php?id=7870
which led me to this: https://cmake.org/Wiki/CMake_MinGW_Compiler_Issues
And here's Appveyor's take: http://help.appveyor.com/discussions/problems/3193-cmake-building-for-mingw-issue-with-git-shexe

### Root Cause

Appveyor starts tests with an environment that includes GIT and everything it needs on your `PATH`.  This includes a copy of `sh.exe`.  Presumably, appveyor needs git to perform the project checkout, and they probably want to keep GIT on the `PATH` in case your project needs to run some other git commands.

The true root cause of the issue is MinGW Make insisting on running commands with `sh` if it's in your `PATH`, which will not work with Makefiles generated using the MinGW Makefile generator. The only way to force MinGW Make to run using `cmd.exe` is to explicitly remove any directory with an executable `sh` from the `PATH`.

### Fix in CMake?

Not wanting to meddle with a user's environment (let alone their `PATH`) seems like a good reason for a CMake maintainer to leave the issue as-is, especially considering that this is only a problem for Windows users who've deliberately added coreutils to their `PATH` (like in appveyor) and the simplest solution in this case is to just use a different generator.

### Fix in Appveyor?

On the other hand, keeping a functional git environment in a user's `PATH` from the outset seems like a more important feature to a member of the Appveyor support staff than catering to the specific needs of CMake users, or anyone who needs to use MinGW over Visual Studio, for that matter.

### What Do?

So, it's not obvious to me who we should be asking to free us from having to work around the issue, ourselves.

appveyor.yml
opadron: This is fine.  The `secure` clause indicates that the string is an encrypted value for which only appveyor has the decryption key.  No sensitive information is being sent in the clear.

ci/appveyor/run-with-visual-studio.cmd
opadron: Found the source.
https://github.com/ogrisel/python-appveyor-demo/blob/master/appveyor/run_with_env.cmd

Will add a comment and push an update, soon.

ci/appveyor/run-with-visual-studio.cmd
opadron: Found the source.
https://github.com/ogrisel/python-appveyor-demo/blob/master/appveyor/run_with_env.cmd

Will add a comment and push an update, soon.

opadron:
### Remaining loose ends

- [ ] Move `sh.exe` removal logic from `appveyor.yml` to `ci/appveyor/run-with-visual-studio.cmd`, to limit the extent of environment meddling only to what is necessary.
- [ ] Update `ci/appveyor/run-with-visual-studio.cmd` with a comment pointing to the original source.
- [ ] Update comment in `ci/appveyor/run-with-visual-studio.cmd` to point to the new location for building extension modules on Windows.
- [ ] Update comment in `appveyor.yml`.

appveyor.yml
opadron: Done.

appveyor.yml
jcfr: Very nice description. Could you add this text to the run-with-mingw.cmd when you will move the "workaroud logic" there ?

ci/appveyor/run-with-visual-studio.cmd
jcfr: Great. Would it makes sense to contribute our improvement and even the file renaming to that repository and just download it from there ?

jcfr: Direct download of specific version could be done using this url: https://raw.githubusercontent.com/ogrisel/python-appveyor-demo/f54ec3593bcea682098a59b560c1850c19746e10/appveyor/run_with_env.cmd

opadron:
@jcfr How does it look, now?

ci/appveyor/run-with-mingw.cmd
```diff
@@ -0,0 +1,23 @@
...
+
+SET COMMAND_TO_RUN=%*
+SET MINGW_ROOT=C:\MinGW
+
+ECHO Using default MinGW build environment
+SET PATH=%MINGW_ROOT%\bin;%PATH%
+
+
+:: workaround for CMake not wanting sh.exe on PATH for MinGW
+
```
jcfr: Could you add the result of investigation as comment here ?

appveyor.yml
```diff
@@ -0,0 +1,164 @@
...
+    # NOTE: For now, Fortran features are not tested #
+    ##################################################
+
+    # Visual Studio (Python 2 & 3, 32 & 64 bit)
+    - PYTHON: "C:\\Python27"
+      PYTHON_VERSION: "2.7.x"
+      PYTHON_ARCH: "32"
+      CMAKE_GENERATOR: Visual Studio 9 2008
+      CMD_IN_ENV: "cmd /E:ON /V:ON /C .\\ci\\appveyor\\run-with-visual-studio.cmd"
+      EXTRA_TEST_ARGS: "--exclude fortran"
```
jcfr: Should this be `--exclude fortran_compiler` ?

See https://github.com/opadron/scikit-build/blob/169667a30776de70cb3529880afdd5ba53c0530f/tests/test_platform.py#L52

opadron:
Huh!? Go home, Appveyor; you're drunk. 👎

opadron:
Looks like Appveyor took a nose dive.  I'll try again later.

jcfr:
Is it a matter of cancelling queued build ?

opadron:
@jcfr no, it's a problem on their end.

<blockquote class="twitter-tweet" data-lang="en"><p lang="en" dir="ltr"><a href="https://twitter.com/nulptrinception">@nulptrinception</a> Looking into that</p>&mdash; AppVeyor CI (@appveyor) <a href="https://twitter.com/appveyor/status/728285273352626177">May 5, 2016</a></blockquote>
<script async src="//platform.twitter.com/widgets.js" charset="utf-8"></script>

opadron:
ready for review, again

appveyor.yml
```diff
@@ -0,0 +1,186 @@
...
+              for this pull request, failing early.\" \
+    }"
+  - cmd: ECHO "Filesystem root:"
+  - ps: "ls \"C:/\""
+  - cmd: ECHO "Installed SDKs:"
+  - ps: "ls \"C:/Program Files/Microsoft SDKs/Windows\""
+
+  # Overwrite the visual-studio environment script with the one from
+  # ogrisel/python-appveyor-demo
+  - ps: "\
```
jcfr: Should we keep `ci/appveyor/run-with-visual-studio.cmd` ?

jcfr:
Nice. Thanks

One small nitpick.

After squashing the commits, LGTM

appveyor.yml
opadron: I think we should.  It certainly doesn't hurt having it around, and we could quickly switch back to using it if we ever need to.

opadron: Although I do see how it might be confusing.  Hmm...I'll take it out, but make sure to preserve it in the history post-squash.

jcfr: Sounds good. 

Exciting to see this integrated.

opadron:
Continued in #32.

PR #32, opened 2016-05-06
[REBASED] CI Integration

opadron:
Rebased version of #31.


opadron:
Funny how Circle CI insists on testing the project even when we haven't added the circle.yml file, yet.

opadron:
Alright I finally got circle ci to accept a dummy circle.yml file.  Since I'd rather not push another rebase branch, I will just rebase in place.

opadron:
Just a quick note to anyone following this: Github seems to be listing the commits in the wrong order on my browser.  If this quirk shows up for others, they will also see the commits in the wrong order.  I thought it may have had something to do with the timestamp on the commit, but that doesn't seem to be the case, either, so I'm not sure how Github is sorting commits.

opadron:
I found this in case anyone's interested: https://help.github.com/articles/why-are-my-commits-in-the-wrong-order.

It's such a shame that Github doesn't even *try* to handle rebases well.

jcfr:
>  the commits in the wrong order on my browser

I am assuming github display them in chronological order ?

Probably equivalent to passing the option `--author-date-order` to `git log`

opadron:
Doesn't seem to make a difference.  Maybe it's some weird interaction with `--topo-order`?

Except not even that makes sense since everything's been rebased to sit in a single branch (that's the whole point).

I think Github is just not sorting.

opadron:
For example, have a look at the commits tab.  There doesn't seem to be any rhyme or reason to the ordering. ¯\\\_(ツ)\_/¯

jcfr:
In the grand scheme ... that is a minor issue.

> just not sorting

Probably .. and doing that sorting when you serve so many million of pages has probably a cost. Rebase being a power user feature, that could explain it.

jcfr:
All of that said, thanks for updating the topic. After the CI comes back green, we are in good shape to merge  then ?

opadron:
almost. one more commit.

opadron:
Final commit is undergoing testing.  If successful, it will be ready to merge.

opadron:
Man, tripped at the finish line.  I know what to do... :)

appveyor.yml
```diff
@@ -106,6 +106,16 @@ install:
   - cmd: ECHO "Installed SDKs:"
   - ps: "ls \"C:/Program Files/Microsoft SDKs/Windows\""
 
+  # Download visual-studio environment script from ogrisel/python-appveyor-demo
+  - ps: "\
+      (New-Object Net.WebClient).DownloadFile(\
+        'https://raw.githubusercontent.com/ogrisel/python-appveyor-demo/\
+          f54ec3593bcea682098a59b560c1850c19746e10/appveyor/run_with_env.cmd', \
+        (
+          Get-Item \".\\ci\\appveyor\\run-with-visual-studio.cmd\" -Verbose\
```
jcfr: Since the file is removed from the repo .. no needed to overwrite it. `Get-Item` can be removed I think.

Then, doing
```
(New-Object Net.WebClient).DownloadFile(https://raw.githubusercontent.com/ogrisel/python-appveyor-demo/f54ec3593bcea682098a59b560c1850c19746e10/appveyor/run_with_env.cmd', 'run-with-visual-studio.cmd')
```
should be enough to download the file in the current dir

jcfr: For this to work, you would then change 

```
CMD_IN_ENV: "cmd /E:ON /V:ON /C .\\ci\\appveyor\\run-with-visual-studio.cmd"
```

to 

```
CMD_IN_ENV: "cmd /E:ON /V:ON /C run-with-visual-studio.cmd"
```

opadron: Unfortunately not, since the full path needs to be passed.  Instead, I just added a dummy file with a comment.

jcfr: The issue is that `Get-Item` will compain with  `ItemNotFound`. 

>  full path needs to be passed.

I don't think it is a requirement from Powershell.

The onliner I copied above is known to work ..

jcfr: Got it. I am slow this morning ... the file `ci/appveyor/run-with-visual-studio.cmd` is a place holder.

opadron: The full path is necessary for the download according to this: http://www.appveyor.com/docs/how-to/download-file.

We could download anew, but I kind of like having a stub in the work tree.  It's a good compromise between not having our own version of the file and letting others know that it's still an important part of the testing.

(That and the tests have just finished successfully, and I really want to merge this and be done with it.) :stuck_out_tongue_closed_eyes:

jcfr: could you add a comment in the appveyor.yml .. mentioning it ?

opadron:
All tests are green.  Ready to merge!

jcfr:
Voila :8ball:

opadron:
👍

PR #36, opened 2016-06-06
CMake Improvements/skbuild features/demo applications

opadron:
So, when I started putting together example applications that were more involved than "hello, world", I soon realized that I needed more tooling improvements to support these new apps.  What started out as an effort to add example applications grew to include CMake scripts as well as several improvements to scikit-build.  I realize this changeset is on the large side, but everything here is related, so I figured breaking things up would obscure more than help.

There's a few technical choices I made while putting this together.  I made them just so that I could make progress, but we should take this time to discuss them in detail and make sure we are all happy with where things are going.  I expect this PR to remain a work in progress for a while as we take the time to work out all the details.

The CMake scripts are improvements that I'd like to see eventually contributed upstream.  This PR should serve as the temporary home until we get those pieces sufficiently polished and ready to propose to the CMake maintainers.

Other than that, I've assembled my notes below.  These notes include a summary of the changes, and a TODO list of things we'll probably want to take care of.

##### scikit-build improvements

 - "_skbuild" directories are now ignored by git.
 - added several changes in how the various staging directories are handled
   - all temporary staging directories are now contained in a top-level directory named "_skbuild"
     - changed the name of the top-level directory from "skbuild" to "_skbuild".
   - Within "_skbuild" are directories with new corresponding constants in `skbuild/cmaker.py`.
     - Location of the CMake project build directory (`cmaker.CMAKE_BUILD_DIR`) is now "_skbuild/cmake-build".
       - changed the name of the CMake project build directory from "cmake" to "cmake-build".
     - Location of the CMake project installation directory (`cmaker.CMAKE_INSTALL_DIR`) is now "_skbuild/cmake-install".
       - changed the constant in `cmaker.py` from `PACKAGE_BUILD_DIR` to `CMAKE_INSTALL_DIR`
       - changed the name of the CMake project installation directory from "lib" to "cmake-install"
     - Location of the package preparation directory used by distutils (`cmaker.DISTUTILS_INSTALL_DIR`) is now "_skbuild/distutils".
     - added custom subclasses of the `build`, `install`, and `clean` commands in order to get distutils to use the new directory.
       - works around a bug in distutils: https://bugs.python.org/issue1011113
         - patches provided, but no updates since 2014
 - skbuild now recognizes the `py_modules` argument and matches entries in the CMake install manifest to modules listed here.
 - skbuild now recognizes the `scripts` argument and matches entries in the CMake install manifest to scripts and executables listed here.
[...]

PR #38, opened 2016-06-09
Testing/Bug Fixes (2016-06-06)

opadron:
Continues the work in #36.

Extends the testing coverage to include the new features and demo applications introduced in #36.

#### TODO

 - [ ] Extend testing coverage to include the new apps and new features.

   - [x] Projects that attempt to CMake-install files outside of its designated project installation root should fail to install.

     - [x] Ideally, they should fail prior to the CMake-install step (i.e.: before any problematic files are actually installed).  This will need code to determine the would-be files without the aid of the install manifest.

   - [x] Each of the sample projects should have their own tests to at least ensure that they build correctly, and that any produced applications run without issue.

    - [x] Add tests for pen2_cython

    - [x] Add tests for tower-of-babel

 - [ ] There are multiple bugs in the way command line arguments are processed and forwarded in scikit-build.  It's not clear if they were introduced by these changes, or if they predate them.

   - [ ] There is a bug when trying to specify the desired CMake build type via the `--build-type` option in scikit-build.  It seems that some sections of the code are looking for `CMAKE_BUILD_DIR/CMAKE_BUILD_TYPE` as the ultimate location of the cmake build directory.  I think handling multiple builds in one tree is desirable and we should probably extend this handling to the other temporary staging directories, but right now, these improvements were not expecting this.  We'll need to go through the various code sections and make sure everyone's on the same page.

   - [ ] On a related note, I tried to pass options directly to CMake through scikit-build using the `--` argument seperator.  Scikit-build is currently written to expect up to three sections of such options: the first for scikit-build options, the second for CMake config options, and the third for options to the build step.  However, since CMake is used to drive the build step, options meant for the actual build tool (e.g.: make) must be separated by another `--` separator.  Currently scikit-build fails to properly handle this fourth separator.

 - [ ] Currently the only way to add to the Cython module search path for `add_cython_target` is to use the outdated `include_directories` CMake command.  Ideally, we would be able to use `target_include_directories` or something similar.  This may involve the creation of an actual CMake target.

### DIFF
--- a/ci/appveyor/run.cmd
+++ b/ci/appveyor/run.cmd
@@ -1,7 +1,7 @@
 
 @SETLOCAL enableextensions enabledelayedexpansion
 
-@ECHO OFF
+:: @ECHO OFF
 
 IF x%SKIP%==x0 (
     IF NOT x%CMAKE_GENERATOR:Visual Studio=%==x%CMAKE_GENERATOR% (
--- a/ci/appveyor/run.cmd
+++ b/ci/appveyor/run.cmd
@@ -3,12 +3,12 @@
 
 :: @ECHO OFF
 
-IF x%SKIP%==x0 (
-    IF NOT x%CMAKE_GENERATOR:Visual Studio=%==x%CMAKE_GENERATOR% (
+IF "x%SKIP%==x0" (
+    IF NOT "x%CMAKE_GENERATOR:Visual Studio=%==x%CMAKE_GENERATOR%" (
         cmd /E:ON /V:ON /C ci\appveyor\run-with-visual-studio.cmd %*
     )
 
-    IF NOT x%CMAKE_GENERATOR:MinGW Makefiles=%==x%CMAKE_GENERATOR% (
+    IF NOT "x%CMAKE_GENERATOR:MinGW Makefiles=%==x%CMAKE_GENERATOR%" (
         cmd /E:ON /V:ON /C ci\appveyor\run-with-mingw.cmd %*
     )
 ) ELSE (
--- a/ci/appveyor/run.cmd
+++ b/ci/appveyor/run.cmd
@@ -1,7 +1,5 @@
-
+@ECHO OFF
 @SETLOCAL enableextensions enabledelayedexpansion
-
-:: @ECHO OFF
 
 IF "x%SKIP%"=="x0" (
     IF NOT "x%CMAKE_GENERATOR:Visual Studio=%"=="x%CMAKE_GENERATOR%" (
@@ -16,4 +14,3 @@ IF "x%SKIP%"=="x0" (
 )
 
 ENDLOCAL
-
--- a/ci/appveyor/run.cmd
+++ b/ci/appveyor/run.cmd
@@ -21,6 +21,6 @@ IF NOT "x%CMAKE_GENERATOR:MinGW=%"=="x%CMAKE_GENERATOR%" ( GOTO mingw )
     echo cmd /E:ON /V:ON /C %script% %*
     cmd /E:ON /V:ON /C %script% %*
 
-done:
+:done
 
 ENDLOCAL
--- a/ci/appveyor/run.cmd
+++ b/ci/appveyor/run.cmd
@@ -2,17 +2,25 @@
 @SETLOCAL enableextensions enabledelayedexpansion
 
 IF "x%SKIP%"=="x0" (
-    IF NOT "x%CMAKE_GENERATOR:Visual Studio=%"=="x%CMAKE_GENERATOR%" (
-        echo cmd /E:ON /V:ON /C ci\appveyor\run-with-visual-studio.cmd %*
-        cmd /E:ON /V:ON /C ci\appveyor\run-with-visual-studio.cmd %*
-    )
-
-    IF NOT "x%CMAKE_GENERATOR:MinGW Makefiles=%"=="x%CMAKE_GENERATOR%" (
-        echo cmd /E:ON /V:ON /C ci\appveyor\run-with-mingw.cmd %*
-        cmd /E:ON /V:ON /C ci\appveyor\run-with-mingw.cmd %*
-    )
-) ELSE (
     ECHO SKIPPING: %*
+    GOTO done
 )
+
+IF NOT DEFINED CMAKE_GENERATOR ( GOTO visual_studio )
+IF NOT "x%CMAKE_GENERATOR:MinGW=%"=="x%CMAKE_GENERATOR%" ( GOTO mingw )
+
+visual_studio:
+    SET script=ci\appveyor\run-with-visual-studio.cmd
+    GOTO run
+
+mingw:
+    SET script=ci\appveyor\run-with-mingw.cmd
+    GOTO run
+
+run:
+    echo cmd /E:ON /V:ON /C %script% %*
+    cmd /E:ON /V:ON /C %script% %*
+
+done:
 
 ENDLOCAL
--- a/ci/appveyor/run.cmd
+++ b/ci/appveyor/run.cmd
@@ -1,7 +1,7 @@
 @ECHO OFF
 @SETLOCAL enableextensions enabledelayedexpansion
 
-IF "x%SKIP%"=="x0" (
+IF "x%SKIP%"=="x1" (
     ECHO SKIPPING: %*
     GOTO done
 )
--- a/ci/appveyor/run.cmd
+++ b/ci/appveyor/run.cmd
@@ -3,12 +3,12 @@
 
 :: @ECHO OFF
 
-IF "x%SKIP%==x0" (
-    IF NOT "x%CMAKE_GENERATOR:Visual Studio=%==x%CMAKE_GENERATOR%" (
+IF "x%SKIP%"=="x0" (
+    IF NOT "x%CMAKE_GENERATOR:Visual Studio=%"=="x%CMAKE_GENERATOR%" (
         cmd /E:ON /V:ON /C ci\appveyor\run-with-visual-studio.cmd %*
     )
 
-    IF NOT "x%CMAKE_GENERATOR:MinGW Makefiles=%==x%CMAKE_GENERATOR%" (
+    IF NOT "x%CMAKE_GENERATOR:MinGW Makefiles=%"=="x%CMAKE_GENERATOR%" (
         cmd /E:ON /V:ON /C ci\appveyor\run-with-mingw.cmd %*
     )
 ) ELSE (
--- a/ci/appveyor/run.cmd
+++ b/ci/appveyor/run.cmd
@@ -6,18 +6,18 @@ IF "x%SKIP%"=="x0" (
     GOTO done
 )
 
-IF NOT DEFINED CMAKE_GENERATOR ( GOTO visual_studio )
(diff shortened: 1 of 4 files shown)


opadron:
Looks like scikit-build can't find a python library on windows?
https://ci.appveyor.com/project/opadron/scikit-build/build/0.0.1.110/job/pmkest6m940n7adr#L273

opadron:
Since there doesn't seem to be much useful platform information from pypy, I say we should drop support for it.

Either that, or we need pypy-specific equivalent for about half a dozen separate sections of code.

cc @jcfr @thewtex

jcfr:
> not much useful platform information from pypy, I say we should drop support for it.

If you are talking about http://pypy.org/ , no need to add support for it at that stage.

tests/samples/tower-of-babel/CMakeLists.txt
```diff
@@ -0,0 +1,77 @@
...
+python_extension_module(tbabel_cython_shared)
+python_extension_module(tbabel_cython_module)
+
+python_extension_module(tbabel_boost_static)
+python_extension_module(tbabel_boost_shared)
+python_extension_module(tbabel_boost_module)
+
+set(site_packages "${PYTHON_RELATIVE_SITE_PACKAGES_DIR}")
+
+INSTALL(TARGETS tbabel_cython_static DESTINATION lib)
```
jcfr: lower case CMake `install(...)`

tests/samples/cmake-files/FindPythonExtensions.cmake
```diff
@@ -222,37 +222,48 @@
...
-set(_command "${_command}        break\n")
-
-set(_command "${_command}sys.stdout.write(\";\".join((\n")
-set(_command "${_command}    os.sep,\n")
-set(_command "${_command}    os.pathsep,\n")
-set(_command "${_command}    sys.prefix,\n")
-set(_command "${_command}    result,\n")
-set(_command "${_command}    rel_result,\n")
-set(_command "${_command})))\n")
+set(_command "
```
jcfr: Should this CMake module be moved into something like `skbuild/cmake` ?

opadron: It's not that important exactly where it lives, since we should be planning to send it upstream when its ready.  I chose `tests/samples/cmake-files` to keep them near the sample projects that would use them.

jcfr: Roger that.

opadron:
**Status update**: Assuming these latest CI runs are successful, we'll be looking at the testing coverage detailed below.  CI testing for the `pen2_cython` and `tower-of-babel` example apps are only partially supported, due to the lack of support in the CI environments for dependencies.

Note that this all assumes that we'd rather avoid tying our CI to Anaconda.  If we are willing to bring in Anaconda, we could improve on the current coverage.

|CI       |Python Version   |toolchain    |pen2_cython       |tower-of-babel    |Notes|
|---------|----------------:|-------------|------------------|------------------|----:|
|Travis   |2.7              |gcc    64-bit|:white_check_mark:|:white_check_mark:|     |
|Travis   |3.3              |gcc    64-bit|:x:               |:x:               | 1, 2|
|Travis   |3.4              |gcc    64-bit|:x:               |:x:               | 1, 2|
|Circle   |2.7              |gcc    64-bit|:x:               |:white_check_mark:|    3|
|AppVeyor |2.7              |VS     32-bit|:x:               |:x:               | 1, 2|
|AppVeyor |2.7              |VS     64-bit|:x:               |:x:               | 1, 2|
|AppVeyor |3.5              |VS     32-bit|:x:               |:x:               | 1, 2|
|AppVeyor |3.5              |VS     64-bit|:x:               |:x:               | 1, 2|
|AppVeyor |2.7              |MinGW  64-bit|:x:               |:x:               | 1, 2|
|AppVeyor |3.5              |MinGW  64-bit|:x:               |:x:               | 1, 2|

###### Notes

  1. Configuration lacks support for tests using VTK; it is either not packaged,
     not compatible with the version of Python, or the platform requires
     additional work to provision a GUI environment.

  1. Configuration lacks support for tests using Boost; it is either not
     packaged, not compatible with the given toolchain or version of Python, or
     does not provide the necessary Python wrapping libraries.

  1. I ran into issues trying to get the python in Circle CI to find the VTK
     bindings upon import.

Considering the poor coverage, I'd like to bring in @jcfr's barebones example, just so that we could have an example app that would be expected to work on all CI environments.

tests/test_hello.py
```diff
@@ -0,0 +1,45 @@
...
+        sys.argv = old_argv
+
+def test_hello_works():
+    old_cwd = os.getcwd()
+    os.chdir(os.path.join("samples", "hello", CMAKE_BUILD_DIR))
+    try:
+        subprocess.check_call(
+            ["ctest", "--build-config",
+                os.environ.get("SKBUILD_CMAKE_CONFIG", "Debug"),
+                "--output-on-failure"])
```
jcfr: Anchor

opadron: I just had another read of pep8, and I'm not sure how this should be formatted.  Feel free to point out the preferred formatting.

opadron:
At this point, I don't think continuing to pursue a working Appveyor CI is the most effective us of our time.  I'd like to suggest that we limit windows testing for the time being so that we can make progress.

We can always come back and continue working on expanding our testing coverage.

What say you, @jcfr?

tests/test_hello.py
jcfr: > Anchor

This was a random text .. to add a comment so that I can reference the text in https://github.com/scikit-build/scikit-build/issues/42

jcfr: No strong feeling about the indent.

jcfr:
I still think having green coverage for the hello world an appveyor makes sense.

Looking at the output associated with https://ci.appveyor.com/project/opadron/scikit-build/build/0.0.1.187/job/749pj42ppjnikaa5

It displays 

```
Environment: PYTHON=C:\Python27-x64, PYTHON_VERSION=2.7.x, PYTHON_ARCH=64, CMAKE_GENERATOR=Visual Studio 9 2008, CMD_IN_ENV=cmd /E:ON /V:ON /C .\ci\appveyor\run-with-visual-studio.cmd, EXTRA_TEST_ARGS=--exclude fortran_compiler --exclude pen2 --exclude tbabel
```

Is it expected to have `PYTHON_ARCH=64` with `Visual Studio 9 2008` ? I would think the generated should be `Visual Studio 9 2008 Win64` instead.

I suspect that updating `appveyor.yml` by associated the `Win64` version of the generator everywhere `C:\\PythonXY-x64` is used should help ...

opadron:
@jcfr good catch! Hopefully, that fixes the problems I've been having!

jcfr:
> Error sending GitHubPullRequest notification: Error commenting on GitHub pull request: {"message":"Bad credentials",

Look like there is also an issue sending back info to github

opadron:
That issue, AFAICT, doesn't affect anything.  I'll see if I can silence it.

opadron:
Closing to try to slap some sense into AppVeyor.

jcfr:
> doesn't affect anything

Great.  If that error message doesn't mean that something is broken, for now  let's keep it and not spend time on it. Let's just just focus on getting the simple hello world green.

opadron:
Had to rebase.  I'm not sure what happened that caused this PR and master to diverge.

opadron:
I can't believe this: http://help.appveyor.com/discussions/kb/38-visual-studio-2008-64-bit-builds

I just want to say this for the record:  "Holy crap, what has been going on with WIndows?"  We are well into Crazy-Town levels of hoop-jumping.  This is insanity.

opadron:
YES!  WOOHOOOO!  I am literally dancing at my desk!
:fireworks: :sparkler: :confetti_ball: :balloon: :fireworks: :sparkler: :confetti_ball: :balloon: :fireworks: :sparkler: :confetti_ball: :balloon: :fireworks: :sparkler: :confetti_ball: :balloon: :fireworks: :sparkler: :confetti_ball: :balloon: :fireworks: :sparkler: :confetti_ball: :balloon: :fireworks: :sparkler: :confetti_ball:

jcfr:
Looking good :8ball:

thewtex:
Congrats!

jcfr:
@opadron Looks like the testing is always disabled .. even when SKIP=0

```
%RUN% %PYTHON%\python.exe .\ci\appveyor\driver.py build
SKIPPING: C:\Python35\python.exe .\ci\appveyor\driver.py build
%RUN% %PYTHON%\python.exe .\ci\appveyor\driver.py test
SKIPPING: C:\Python35\python.exe .\ci\appveyor\driver.py test
%RUN% %PYTHON%\python.exe .\ci\appveyor\driver.py after_test
SKIPPING: C:\Python35\python.exe .\ci\appveyor\driver.py after_test
Packaging artifacts...OK
```

jcfr:
Looking at https://github.com/opadron/scikit-build/blob/6e6ff22fe68309e8ce9f7e478a548d1da4d23c3b/ci/appveyor/run.cmd#L4-L7

Should `"x%SKIP%"=="x0"` be changed to `"x%SKIP%"=="x1"` ?

opadron:
Of course.  Can't believe I missed that. :(

jcfr:
Looking good. Almost all the tests pass on Appveyor :+1:   Two left ... 

After squashing the commit .. we should be in good shape for integration

opadron:
Unfortunately, the build errors are being masked, it seems.

https://ci.appveyor.com/project/opadron/scikit-build/build/0.0.1.243/job/u3fwr57i2p8xxtvq#L218

jcfr:
> the build errors are being masked

See http://stackoverflow.com/questions/734598/how-do-i-make-a-batch-file-terminate-upon-encountering-an-error

I suspect error code is not propagated from one script to an other.

ci/appveyor/run.cmd
```diff
@@ -19,7 +19,7 @@ IF NOT "x%CMAKE_GENERATOR:MinGW=%"=="x%CMAKE_GENERATOR%" ( GOTO mingw )
 
 :run
     echo cmd /E:ON /V:ON /C %script% %*
-    cmd /E:ON /V:ON /C %script% %*
+    cmd /E:ON /V:ON /C %script% %* || EXIT 1
```
jcfr: Does this mean it will always return 1 ?

opadron: No, this is only in the case of failure.

jcfr: Great.  Makes sense.

```
$ bash -c "echo 'failure';exit 1" || echo "propagate"
failure
propagate

$ bash -c "echo 'success';exit 0" || echo "nothing"
success
```

jcfr:
Nicee. That is worth an other dance on the desk :smile:

jcfr:
Do you know why the last two are skipped ? Look like they should be passing .. assuming the auto selection of the generator work as expected. May be you need to make sure the `Win64` suffix is added to the generator ?

opadron:
@jcfr Yeah, I already made a note of that problem.  I know where it is, too, but we need to move on.  I'll add everything I know to the issue tracker so that we won't forget.

jcfr:
No problems. Makes sense.

opadron:
superseded in #43

PR #41, opened 2016-06-15
fix #40 - tweak visual studio generator selection

### LINKED ISSUE #40: Use PyPA recommended VS compilers
The PyPA has a guide line one the versions of compilers to use with difference versions of Python [1].

Specifically, I noticed that 2.7 was correctly compiled with VS9 2008, while py 3.5 was incorrectly compiled with VS10 2010.  From the document:

> Python 2.7 used Visual Studio 2008, Python 3.3 and 3.4 used Visual Studio 2010, and Python 3.5+ uses Visual Studio 2015.

[1] https://media.readthedocs.org/pdf/python-packaging-user-guide/develop/python-packaging-user-guide.pdf

opadron:


opadron:
Assuming AppVeyor is happy, I suggest we merge in spit of CircleCI.  The issue with circle is already known and being addressed in #38

opadron:
@scikit-build/developers ☑️  Ready!

jcfr:
Thanks

PR #47, opened 2016-07-06
Weak linking CMake module

opadron:
Adapted from [vtk#1511](https://gitlab.kitware.com/vtk/vtk/merge_requests/1511).

Adds a new experimental CMake module: `targetLinkLibrariesWithDynamicLookup` that allows one to portably indicate that a target uses symbols from another library, but that the resolution of symbols should be delayed until runtime, i.e.: the linker should tolerate undefined symbols when linking.  "Weak linking" has been used to refer to this code sharing scheme as well as its many subtle variants.

Reviewers, please note the major differences in this PR when compared to [vtk#1511](https://gitlab.kitware.com/vtk/vtk/merge_requests/1511).
  - This version uses two instances of a sample project with `try_compile` to infer the level of support for dynamic symbol resolution.  The first determines if the linker responds to the dynamic lookup flag currently present in linker on OS X.  The second determines if the linker and run-time loader can properly coalesce symbols that have been duplicated across link boundaries, such as `ld`/Linux.  The original version only included the first check.
  - In addition, each sample project is tested after compiling to ensure that the test program produces the expected output as well as compile successfully.  The original used a much simpler test program that would only compile.
  - In the original version, no link is performed when the linker is found to not support the dynamic lookup flag.  In general, a proper linking is still necessary on platforms like Linux, whose loader can coalesce duplicate symbols at run-time.  In this version, the link is performed when such a loader has been detected.
  - The original version's system checks were performed immediately upon module load and then cached.  This version runs the system checks and caches on demand. 

Also, includes a fix for when testing the `tower-of-babel` sample project under anaconda.

### DIFF
--- a/tests/samples/cmake-files/targetLinkLibrariesWithDynamicLookup.cmake
+++ b/tests/samples/cmake-files/targetLinkLibrariesWithDynamicLookup.cmake
@@ -102,7 +102,10 @@
 # weak-linking, this function works just like ``target_link_libraries``.
 
 function(_get_target_type result_var target)
-  get_property(target_type TARGET ${target} PROPERTY TYPE)
+  set(target_type "SHARED_LIBRARY")
+  if(TARGET ${target})
+    get_property(target_type TARGET ${target} PROPERTY TYPE)
+  endif()
 
   set(result "STATIC")
 
--- a/tests/samples/cmake-files/targetLinkLibrariesWithDynamicLookup.cmake
+++ b/tests/samples/cmake-files/targetLinkLibrariesWithDynamicLookup.cmake
@@ -1,7 +1,7 @@
 #
 # - This module provides the function
 # target_link_libraries_with_dynamic_lookup which can be used to
-# "weakly" link loadable module.
+# "weakly" link a loadable module.
 #
 # Link a library to a target such that the symbols are resolved at
 # run-time not link-time. This should be used when compiling a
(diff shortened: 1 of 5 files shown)


tests/samples/cmake-files/targetLinkLibrariesWithDynamicLookup.cmake
```diff
@@ -0,0 +1,278 @@
...
+              CMAKE_FLAGS
+                "-DCMAKE_SHARED_LINKER_FLAGS='${CMAKE_SHARED_LINKER_FLAGS}'"
+                ${_rpath_arg}
+              OUTPUT_VARIABLE compile_output)
+
+  set(project_works 1)
+  set(run_output)
+
+  if(project_compiles)
+    execute_process(COMMAND "${test_project_bin_dir}/main"
```
jcfr:
```
execute_process(COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} "${test_project_bin_dir}/main"
```

tests/samples/cmake-files/targetLinkLibrariesWithDynamicLookup.cmake
```diff
@@ -0,0 +1,278 @@
...
+endfunction()
+
+
+function(has_dynamic_lookup resultVar)
+  # hash the CMAKE_FLAGS passed and check cache to know if we need to rerun
+  string(MD5 cmake_flags_hash "${CMAKE_SHARED_LINKER_FLAGS}")
+
+  if(     NOT DEFINED HAS_DYNAMIC_LOOKUP_hash
+       OR NOT "${HAS_DYNAMIC_LOOKUP_hash}" STREQUAL "${cmake_flags_hash}")
+    unset(HAS_DYNAMIC_LOOKUP)
```
jcfr:
```
unset(HAS_DYNAMIC_LOOKUP CACHE)
```

jcfr: Should `HAS_DYNAMIC_LOOKUP_hash` be unset too ?

tests/samples/cmake-files/targetLinkLibrariesWithDynamicLookup.cmake
```diff
@@ -0,0 +1,278 @@
...
+endfunction()
+
+
+function(has_symbol_dedupe resultVar)
+  # hash the CMAKE_FLAGS passed and check cache to know if we need to rerun
+  string(MD5 cmake_flags_hash "${CMAKE_SHARED_LINKER_FLAGS}")
+
+  if(     NOT DEFINED HAS_SYMBOL_DEDUPE_hash
+       OR NOT "${HAS_SYMBOL_DEDUPE_hash}" STREQUAL "${cmake_flags_hash}")
+    unset(HAS_SYMBOL_DEDUPE)
```
jcfr:
```
unset(HAS_SYMBOL_DEDUPE CACHE)
```

Should `HAS_SYMBOL_DEDUPE_hash` be unset ?

tests/samples/cmake-files/targetLinkLibrariesWithDynamicLookup.cmake
```diff
@@ -0,0 +1,278 @@
...
+#   Whether the linker/loader correctly coalesces duplicated symbols
+#
+# ``HAS_SYMBOL_DEDUPE``
+#   Cached alias.
+#
+
+function(_test_weak_link_project projectName testName)
+  set(options DYNAMIC_LOOKUP SYMBOL_DEDUPE)
+  cmake_parse_arguments(_args "${options}" "" "" ${ARGN})
+
```
jcfr:
```
# Sanity check
if(_args_DYNAMIC_LOOKUP AND _args_SYMBOL_DEDUPE)
  message(FATAL_ERROR "Option DYNAMIC_LOOKUP and SYMBOL_DEDUPE are mutually exclusive")
endif()
```

tests/samples/cmake-files/targetLinkLibrariesWithDynamicLookup.cmake
```diff
@@ -0,0 +1,278 @@
...
+  set(test_project_src_dir
+      "${PROJECT_BINARY_DIR}/CMakeTmp/${projectName}/src")
+  set(test_project_bin_dir
+      "${PROJECT_BINARY_DIR}/CMakeTmp/${projectName}/build")
+
+  file(MAKE_DIRECTORY ${test_project_src_dir})
+  file(MAKE_DIRECTORY ${test_project_bin_dir})
```
jcfr: Could you include the description of the test project ? 

May be by adapting/copying part of the description that is available here: https://github.com/opadron/weak-linking-demo ?

tests/samples/cmake-files/targetLinkLibrariesWithDynamicLookup.cmake
```diff
@@ -0,0 +1,278 @@
...
+              "${projectName}"
+              CMAKE_FLAGS
+                "-DCMAKE_SHARED_LINKER_FLAGS='${CMAKE_SHARED_LINKER_FLAGS}'"
+                ${_rpath_arg}
+              OUTPUT_VARIABLE compile_output)
+
+  set(project_works 1)
+  set(run_output)
+
+  if(project_compiles)
```
jcfr:
```
set(do_try_run ${project_compiles})
if(do_try_run AND CMAKE_CROSSCOMPILING)
  if(NOT DEFINED CMAKE_CROSSCOMPILING_EMULATOR)
    set(do_try_run FALSE)
  else()
    set(project_works TRUE)
  endif()
endif()
if(do_try_run)
```

jcfr: We should implement the same behaviour as `try_run `.... 
See https://cmake.org/cmake/help/git-next/command/try_run.html#behavior-when-cross-compiling

tests/samples/cmake-files/targetLinkLibrariesWithDynamicLookup.cmake
```diff
@@ -0,0 +1,278 @@
...
+       OR NOT "${HAS_DYNAMIC_LOOKUP_hash}" STREQUAL "${cmake_flags_hash}")
+    unset(HAS_DYNAMIC_LOOKUP)
+  endif()
+
+  if(NOT DEFINED HAS_DYNAMIC_LOOKUP)
+    _test_weak_link_project(has_dynamic_lookup
+                            HAS_DYNAMIC_LOOKUP
+                            DYNAMIC_LOOKUP)
+
+    set(HAS_DYNAMIC_LOOKUP ${has_dynamic_lookup}
```
jcfr: Should the cache var be named `LINKER_SUPPORT_DYNAMIC_LOOKUP` ?

tests/samples/cmake-files/targetLinkLibrariesWithDynamicLookup.cmake
```diff
@@ -0,0 +1,278 @@
...
+       OR NOT "${HAS_SYMBOL_DEDUPE_hash}" STREQUAL "${cmake_flags_hash}")
+    unset(HAS_SYMBOL_DEDUPE)
+  endif()
+
+  if(NOT DEFINED HAS_SYMBOL_DEDUPE)
+    _test_weak_link_project(has_symbol_dedupe
+                            HAS_SYMBOL_DEDUPE
+                            SYMBOL_DEDUPE)
+
+    set(HAS_SYMBOL_DEDUPE ${has_symbol_dedupe}
```
jcfr: Should the cache var be named `LINKER_SUPPORT_SYMBOL_DEDUPE` ? Or  `LINKER_HAS_SYMBOL_DEDUPE`

tests/samples/cmake-files/targetLinkLibrariesWithDynamicLookup.cmake
```diff
@@ -0,0 +1,278 @@
...
+endfunction()
+
+
+function(target_link_libraries_with_dynamic_lookup target)
+  has_dynamic_lookup(_dlookup)
+  if(_dlookup)
+    set_target_properties(${target}
+                          PROPERTIES LINK_FLAGS "-undefined dynamic_lookup")
+  else()
+    has_symbol_dedupe(_dedupe)
```
jcfr: Is this test required ? Can we safely assume it is already the case ?

blowekamp:
"In the original version, no link is performed when the linker is found to not support the dynamic lookup flag. In general, a proper linking is still necessary on platforms like Linux, whose loader can coalesce duplicate symbols at run-time. In this version, the link is performed when such a loader has been detected."

The goal of the original cmake file (in SimpleITK) was NOT to link again libpython on Linux and OSX as per distutils and Pypa recommendations. This sounds like you are trying to still link against the library?

I am not sure what the goal of this is now...

tests/samples/cmake-files/targetLinkLibrariesWithDynamicLookup.cmake
```diff
@@ -0,0 +1,278 @@
...
+
+    set(HAS_SYMBOL_DEDUPE_hash "${cmake_flags_hash}"
+        CACHE INTERNAL "hashed flags for HAS_SYMBOL_DEDUPE check")
+  endif()
+
+  set(${resultVar} "${HAS_SYMBOL_DEDUPE}" PARENT_SCOPE)
+endfunction()
+
+
+function(target_link_libraries_with_dynamic_lookup target)
```
blowekamp: This is the original logic:
 ```
if ( ${CMAKE_SYSTEM_NAME} MATCHES "Darwin" )
     set_target_properties( ${target} PROPERTIES LINK_FLAGS "-undefined dynamic_lookup" )
   elseif(VTK_UNDEFINED_SYMBOLS_ALLOWED)
     # linker allows undefined symbols, let's just not link
   else()
     target_link_libraries ( ${target} ${ARGN}  )
   endif()
```

This is doing something different now.

**Comment by blowekamp** [comment]
I am also going to point out the having a single expectable linked to two shared libraries, has quite a different scope for symbol resolution that an executable dynamically loading at runtime two separate libraries.

**Comment by jcfr** [comment]
Thanks for the review brad.

For references:
* https://www.python.org/dev/peps/pep-0513/#libpythonx-y-so-1
* [pypa/manylinux/issues/30#issuecomment-181208636](https://github.com/pypa/manylinux/issues/30#issuecomment-181208636)
* [pypa/manylinux/pull/31](https://github.com/pypa/manylinux/pull/31/commits)

**Comment by blowekamp** [comment]
This I think is how I began down this issue by naively linking against the shared library:
* https://github.com/pypa/manylinux/issues/39
* https://mail.python.org/pipermail/distutils-sig/2016-January/028079.html

**Comment by opadron** [comment]
The logic change was on purpose, but clearly, it was a mistake if everything I read is true.

 > `# linker allows undefined symbols, let's just not link`

So, my main confusion stems from the fact that I could not replicate this behavior on Linux -- the linker always refuses to leave the symbols undefined.  Is there a linker flag I should be using?

 > I am also going to point out that having a single executable linked to two shared libraries has quite a different scope for symbol resolution than an executable dynamically loading at runtime two separate libraries.

This sounds right to me, but I'm not sure what the point of bringing this up is?  Neither of those scenarios are being tested in the test project.  Should they be?  Besides this tolerance for undefined symbols that I have yet to observe, is there anything we should be checking for that we're not?

**Comment by blowekamp** [comment]
What is your platform and what exactly are you trying to do to test if "undefined" symbols are allowed? The original CMake file seems to pass on all my linux (elf binaries) systems I have tests [1].

I may have been mistaken about what this CMake code and the test project was trying to do. And I don't know why either.

I also don't know what problem or issue the HAS_SYMBOL_DEDUPE is trying to address.

I thought I was understanding the linker process pretty well now, but I am guess I have more to read.

[1] https://github.com/SimpleITK/SimpleITK/blob/master/CMake/sitkTargetLinkLibrariesWithDynamicLookup.cmake#L44-L64

**Comment by opadron** [comment]
I haven't actually tried that particular test code, but I was more referring to building actual Python extension modules.  When trying to build a simple extension module, the link fails with undefined symbol errors on Linux.  That is- when I don't link to any libpython.

The test project in this PR is based on my experiments [here](https://github.com/opadron/weak-linking-demo).  There is a shared lib, L; a module lib, M; and the executable, E.  E is linked against L.  E also dynamically loads M.  And M, itself uses code from L, but the main question is whether M should be linked against L.  This test was supposed to emulate a typical example of building a python extension module.  As an aside, I haven't been able to build M without linking it against L on Linux, I run into the exact same problems as I did with actual Python extension modules.

I figure besides library version mismatches, the only other potential source of problems with linking for both the module and the executable is that it may result in code duplication.  If the library had some shared state, we should be able to observe the effects of such duplication by inspecting the shared state from the perspectives of E and M, and then comparing.  If there is code duplication, the counting routines from each would have their own counters instead of sharing.  If the loader coalesces the duplicate symbols, then they would share, which is the desired behavior.  This is what I check for under the "SYMBOL_DEDUPING" name.

I'm going to check to see if I can get your simpler test code to compile and link.  If so, then there's clearly something about how linking on Linux works that I'm not privy to.

**Comment by blowekamp** [comment]
Another specific case where not linking solved a problem was on OS X. With binaries of SimpleITK compiled against the system Python 2.7 with explicit linking, when these would create a load error when run on Anaconda or home brew. However, if the "-undefined dynamic_lookup"  was used and not linked against libpython, then the binaries would work!

SimpleITK is not linking against libpython, on every linux system I have tested RH5,RH6,RH7, Cent OS5,  Some Debian, Ubuntu 12, Ubuntu 14....

I'd be interested in a minimal example and the link line used. I'm thinking of starting tests with gcc command lines here.

**Comment by opadron** [comment]
Well, I've finally gotten extension modules to build and link on Linux as you and others describe -- without linking.

I have no idea what I'm doing differently, but I'm now a believer. :sweat_smile: 

I want to try to incorporate this into my weak-linking-demo and see how that changes things.  Then, I'll come back to this and make the necessary corrections.

Thanks for keeping me honest. :+1:

**Comment by opadron** [comment]
Check out opadron/weak-linking-demo#3 for an updated version of the module and an updated discussion in the README.

**Comment by opadron** [comment]
This should be ready to go.

File: tests/samples/cmake-files/targetLinkLibrariesWithDynamicLookup.cmake [hunk 11]
```diff
@@ -0,0 +1,470 @@
...
+        counter_module = dlopen(\"./counter.so\", RTLD_LAZY | RTLD_GLOBAL);
+        if(!counter_module) goto error;
+
+        count = dlsym(counter_module, \"count\");
+        if(!count) goto error;
+      ")
+    endif()
+
+    file(APPEND "${test_project_src_dir}/main.c" "
+        result = count()    != 0 ? 1 :
```
jcfr: `1` -> `EXIT_FAILURE`

tests/samples/cmake-files/targetLinkLibrariesWithDynamicLookup.cmake
```diff
@@ -0,0 +1,470 @@
...
+  set(cache_hash_var "HAS_DYNAMIC_LOOKUP_${target_type}_${lib_type}_hash")
+  set(result_var "DYNAMIC_LOOKUP_FLAGS_${target_type}_${lib_type}")
+
+  if(     NOT DEFINED ${cache_hash_var}
+       OR NOT "${${cache_hash_var}}" STREQUAL "${cmake_flags_hash}")
+    unset(${cache_var} CACHE)
+  endif()
+
+  if(NOT DEFINED ${cache_var})
+    if(NOT CMAKE_CROSSCOMPILING OR CMAKE_CROSSCOMPILING_EMULATOR)
```
jcfr: `if(NOT CMAKE_CROSSCOMPILING OR (CMAKE_CROSSCOMPILING AND CMAKE_CROSSCOMPILING_EMULATOR))`

opadron:
continued in #51

PR #52, opened 2016-07-16
add CMake variable to project builds: "SKBUILD"

opadron:
Adds a new CLI-defined variable to CMake builds that are ran by scikit-build: "SKBUILD".
Set to "TRUE" when so ran.

### DIFF
--- a/skbuild/cmaker.py
+++ b/skbuild/cmaker.py
@@ -241,7 +241,7 @@ class CMaker(object):
                '-DPYTHON_VERSION_STRING=' + sys.version.split(' ')[0],
                '-DPYTHON_INCLUDE_DIR=' + python_include_dir,
                '-DPYTHON_LIBRARY=' + python_library,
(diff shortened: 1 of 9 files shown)


skbuild/cmaker.py
```diff
@@ -241,6 +241,7 @@ def configure(self, clargs=(), generator_id=None):
                '-DPYTHON_VERSION_STRING=' + sys.version.split(' ')[0],
                '-DPYTHON_INCLUDE_DIR=' + python_include_dir,
                '-DPYTHON_LIBRARY=' + python_library,
+               '-DSKBUILD=TRUE',
```
thewtex: -DSKBUILD:BOOL=TRUE

to ensure the correct type.

opadron:
@scikit-build/developers PTAL

jcfr:
Nice.

Could you squash ? Thanks

opadron:
#60

PR #76, opened 2016-07-18
Patches to get the tests passing locally and use types for CMake variables

thewtex:


jcfr:
I am integrating. Appveyor failure has been fixed by #77

Issue #85, opened 2016-07-18
Fix python setup.py build bdist_wheel


thewtex:
Currently outputs::

  usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
   or: setup.py --help [cmd1 cmd2 ...]
   or: setup.py --help-commands
   or: setup.py cmd --help

  error: invalid command 'bdist_wheel'

opadron:
#89

jcfr:
Cc: @opadron @msmolens @thewtex  

To keep clarify in the history of the code, I am reorganizing the commits that were integrated as of https://github.com/scikit-build/scikit-build/pull/89 and will push force. Noise (addition of change, partial revert, ... ) was introduced in critical part of the code by the sequences of commits. This makes it hard to review and track the changes.

When done, you will have to `rebase --onto` of your currently open topics.

Issue #63, opened 2016-07-17
Refactor testing of scikit-build sample projects


jcfr:
Currently [ctest is invoked](https://github.com/scikit-build/scikit-build/blob/26dae334d705805d95ac6d99571e0284fd818a09/tests/test_pen2.py#L39-L42) to run test of each sample projects and we should also [avoid explicitly setting](https://github.com/scikit-build/scikit-build/blob/26dae334d705805d95ac6d99571e0284fd818a09/tests/samples/pen2-cython/scripts/CMakeLists.txt#L14) `PYTHONPATH`. Mocking with `PYTHONPATH` is causing problem with tools like a `tox`.

Instead, sample project level tests:
* should be implemented as regular python test using either `node` or python `unittest`.
* sample project tests should be executed in two contexts:
 * developer mode / editable package mode
 * installed package mode

jcfr:
xref #42

jcfr:
As a side note, tox works great:

```
$ tox
[...]
  py27: commands succeeded
ERROR:   py33: InterpreterNotFound: python3.3
  py34: commands succeeded
  py35: commands succeeded

```

henryiii:
scikit-build (classic) after #1185 is now a thin wrapper around scikit-build-core's setuptools plugin. Closing all open issues; if this still applies, please reopen or open an issue on scikit-build-core. Thanks!

PR #89, opened 2016-07-20
Switch to setuptools

opadron:
Building wheels should now work!

Also refactors some of the test code.  Subproject tests (defined in their `CMakeLists.txt` files) are currently commented out.  They will later be replaced with nose-friendly versions.

### DIFF
--- a/skbuild/cmaker.py
+++ b/skbuild/cmaker.py
@@ -110,37 +110,23 @@ class CMaker(object):
         python_library = CMaker.get_python_library(python_version)
 
         cwd = os.getcwd()
-        cmd = ['cmake', cwd, '-G', generator_id]
-        cmd.extend(
-            "=".join((cmake_variable, value)) for cmake_variable, value in (
-                (
-                    "-DCMAKE_INSTALL_PREFIX:PATH",
-                    os.path.join(cwd, CMAKE_INSTALL_DIR)
-                ), (
-                    "-DPYTHON_EXECUTABLE:FILEPATH",
-                    sys.executable
-                ), (
-                    "-DPYTHON_VERSION_STRING:STRING",
-                    sys.version.split(' ')[0]
-                ), (
-                    "-DPYTHON_INCLUDE_DIR:PATH",
-                    python_include_dir
-                ), (
-                    "-DPYTHON_LIBRARY:FILEPATH",
-                    python_library
-                ), (
-                    "-DSKBUILD:BOOL",
-                    "TRUE"
-                ), (
-                    "-DCMAKE_MODULE_PATH:PATH",
-                    os.path.join(
-                        os.path.dirname(__file__),
-                        "resources",
-                        "cmake"
-                    )
-                )
-            )
-        )
+        cmd = [
+            'cmake', cwd, '-G', generator_id,
+            ("-DCMAKE_INSTALL_PREFIX:PATH=" +
+                os.path.join(cwd, CMAKE_INSTALL_DIR)),
+            ("-DPYTHON_EXECUTABLE:FILEPATH=" +
+                sys.executable),
+            ("-DPYTHON_VERSION_STRING:STRING=" +
+                sys.version.split(' ')[0]),
+            ("-DPYTHON_INCLUDE_DIR:PATH=" +
+                python_include_dir),
+            ("-DPYTHON_LIBRARY:FILEPATH=" +
+                python_library),
+            ("-DSKBUILD:BOOL=" +
+                "TRUE"),
+            ("-DCMAKE_MODULE_PATH:PATH=" +
+                os.path.join(os.path.dirname(__file__), "resources", "cmake"))
+        ]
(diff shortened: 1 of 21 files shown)


opadron:
This should fix #85

thewtex:
woot!  LGTM, but CI is not yet passing

opadron:
@thewtex could you take another pass at this?  I've added some more changes to get CI working.

skbuild/cmaker.py
```diff
@@ -103,24 +102,45 @@ def configure(self, clargs=(), generator_id=None):
...
-               '-DPYTHON_INCLUDE_DIR:PATH=' + python_include_dir,
-               '-DPYTHON_LIBRARY:FILEPATH=' + python_library,
-               '-DSKBUILD:BOOL=TRUE',
-               "-DCMAKE_MODULE_PATH:PATH={}".format(
-                   os.path.dirname(__file__) + '/resources/cmake')
-               ]
+        cwd = os.getcwd()
+        cmd = ['cmake', cwd, '-G', generator_id]
+        cmd.extend(
+            "=".join((cmake_variable, value)) for cmake_variable, value in (
```
jcfr: Could we simply extend the list, I don't think using the join and for loop give us much here ?

skbuild/cmaker.py
```diff
@@ -151,27 +171,61 @@ def get_python_version():
...
-                os.path.join(sysconfig.get_python_inc(),
-                             ".".join(map(str, sys.version_info[:2]))),
-                sysconfig.get_python_inc()
-                ]
+        if not found_python_h:
+            candidate_prefixes = []
+
+            # NOTE(opadron): these possible prefixes must be guarded against
+            # AttributeErrors and KeyErrors because they each can throw on
+            # different platforms or even different builds on the same platform.
```
jcfr: Since the documentation indicate `None` is returned, could you revert these changes ?

* https://docs.python.org/2/library/sysconfig.html#sysconfig.get_config_var
* https://docs.python.org/3/library/sysconfig.html#sysconfig.get_config_var

On Linux, no error:
```
$ python -c "import sysconfig; print(sysconfig.get_config_var('PYTHONPATH'))"
:plat-x86_64-linux-gnu:lib-tk:lib-old

$ python -c "import sysconfig; print(sysconfig.get_config_var('Invalid'))"
None
```

On MacOSx, no error:
```
$ python -c "import sysconfig; print(sysconfig.get_config_var('Invalid'))"
None
$ python -c "import sysconfig; print(sysconfig.get_config_var('PYTHONPATH'))"
:plat-darwin:plat-mac:plat-mac/lib-scriptpackages:../../Extras/lib/python:lib-tk:lib-old
```

On windows, I expect the same.

skbuild/setuptools_wrap.py
```diff
@@ -59,19 +63,21 @@ def parse_args():
                              concatenate_value=True)
     dutils, cmake = move_arg('-G', dutils, cmake)
     dutils, make = move_arg('-j', dutils, make)
-    op = os.path
-
-    def absappend(x):
-        return op.join(op.dirname(op.abspath(sys.argv[0])), x)
+    absappend = (
+        lambda x: os.path.join(
```
jcfr: Do not use lambda here. See https://github.com/scikit-build/scikit-build/commit/1cd88e7b24cfe03fb06472de8a6619cc38f96439

jcfr:
xref #63 (Refactor testing of scikit-build sample projects)

skbuild/cmaker.py
opadron: Different strokes, I guess.  I had hoped this would make the arguments that we're passing easier to read through.

skbuild/cmaker.py
opadron: You're right, that particular piece does not need to be guarded.

skbuild/setuptools_wrap.py
opadron: Any particular reason?

blowekamp: JC provided a link in the patch linked:
http://stackoverflow.com/questions/25010167/e731-do-not-assign-a-lambda-expression-use-a-def

PEP-8

opadron: @blowekamp I see, now.  I didn't notice that link.  Makes sense to me.

opadron:
@scikit-build/developers PTAL

thewtex:
:+1:

opadron:
Wheel builders, rejoice! 🎉

jcfr:
It turns out that this topic broke the submission of `coverage.xml`, it is now generated in this folder:

```
./tests/samples/fail-outside-project-root/coverage.xml
```

instead of 

```
./tests/coverage.xml
```

jcfr:
It turns out that the test `outside_project_root` is causing the `coverage.xml` to be generated in "wrong" location

PR #99, opened 2016-07-21
documentation: Fix formatting of requirements

msmolens:
This commit fixes the formatting of the Python package dependencies in the
documentation by including the requirements*.txt files as literal blocks.

Reference:
http://docutils.sourceforge.net/docs/ref/rst/directives.html#including-an-external-document-fragment


jcfr:
Thanks for the patch

PR #104, opened 2016-07-22
Simplify testing and fix coverage

jcfr:


tests/test_outside_project_root.py
```diff
@@ -10,20 +10,23 @@
...
-
-    exception_thrown = False
-    try:
-        should_fail()
-    except SKBuildError:
-        exception_thrown = True
-
-    assert exception_thrown
+
+    with push_dir():
```
opadron: This shouldn't be necessary, since the `project_setup_py_test` decorator should already take care of that.

jcfr: Without this, the `coverage.xml` file is generated in the test subdirectory.

thewtex:
:+1:

jcfr:
@thewtex @opadron   Thanks for review and integration

PR #107, opened 2016-07-25
[WIP] Fix location of package prefixes

thewtex:
The current implementation had logic errors -- it did not support paths containing a
'.'.

Support the "root package" and duplicate the logic from distutils, which is
documented and expected.

  https://docs.python.org/2/distutils/setupscript.html#listing-whole-packages

Note that this commit starts a new test directory, `unit`.

### DIFF
--- a/skbuild/setuptools_wrap.py
+++ b/skbuild/setuptools_wrap.py
@@ -11,10 +11,7 @@ from . import cmaker
 from .command import build, install, clean, bdist, bdist_wheel, egg_info
 from .exceptions import SKBuildError
 
-try:
-    from setuptools import setup as upstream_setup
-except ImportError:
-    from distutils.core import setup as upstream_setup
+from setuptools import setup as upstream_setup
 
 
 def create_skbuild_argparser():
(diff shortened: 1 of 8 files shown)


thewtex:
Rebased on upstream/master

thewtex:
I will work on improving the coverage...

jcfr:
> I will work on improving the coverage...

Thanks. That would be nice.

skbuild/setuptools_wrap.py
```diff
@@ -194,7 +236,7 @@ def _classify_files(install_paths, package_data, package_prefixes, py_modules,
                 "    Project Root  : {}\n"
                 "    Violating File: {}\n").format(install_root, test_path))
 
-        # peel off the 'skbuild' prefix
+            # peel off the 'skbuild' prefix
```
opadron: Should be aligned with code on the next line.

tests/unit/root-package/lib/_hello.cxx
```diff
@@ -0,0 +1,65 @@
...
+  "Internal \"_hello\" module",
+  -1,
+  hello_methods
+};
+
+PyMODINIT_FUNC PyInit__hello(void)
+{
+  return PyModule_Create(&hello_module_def);
+}
+#endif /* PY_MAJOR_VERSION >= 3 */
```
opadron: Could we have a much more stripped down module example?  We should already have this code in the `hello` test project, and I'd rather not duplicate code needlessly.

skbuild/setuptools_wrap.py
thewtex: Done

tests/unit/root-package/lib/_hello.cxx
thewtex: I will update with a spam example.

thewtex:
:-1: Please do not merge. Do possibly to a subtle change in setuptools (?), the wheel generated by the test no longer contains the C extension.

Closing because I don't think it is worth putting more effort into this path. Focusing instead on #124

PR #109, opened 2016-07-25
tests: Do not call both include(CTest) and enable_testing()

thewtex:
They are redundant and only one or the other should be called per best
practices.


jcfr:
Thanks. Makes sense.

That said, I think they are not exactly redundant:
* call to `enable_testing()`: Allow use of `add_test` 
* include of `CTest`: implies `enable_testing` + enable CTest and dashboard submissions

Could you adjust the commit message accordingly ?

thewtex:
> Could you adjust the commit message accordingly ? 

Thanks for the review.  Added.

PR #114, opened 2016-07-29
Switch to py.test testing tool

### LINKED ISSUE #91: RFC: Transition to py.test
The `py.test` framework integrates better with setuptools.

See http://docs.pytest.org/en/latest/goodpractices.html#integrating-with-setuptools-python-setup-py-test-pytest-runner

It will also allow to remove workaround discussed in #90

msmolens:
Because nose is not being maintained, switch from nose to py.test
(http://docs.pytest.org/).

From https://nose.readthedocs.io/en/latest/:

Nose has been in maintenance mode for the past several years and will likely
cease without a new person/team to take over maintainership. New projects should
consider using Nose2, py.test, or just plain unittest/unittest2.

Fixes #91

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -2,3 +2,6 @@
 branch = True
 source = skbuild
 include = */skbuild/*
+
+[xml]
+output = tests/coverage.xml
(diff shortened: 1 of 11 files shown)


thewtex:
:+1: Nice!

jcfr:
:+1:

Issue #57, opened 2016-07-16
Fix arg parsing


opadron:
Same as #56, but we need to fix the CLI arg parsing logic.

jcfr:
CMake option should exposed using an interface similar to what is done in symengine.py. See https://github.com/symengine/symengine.py/blob/b5d63ebd901030d668b717aeeb8547aa5a14726c/setup.py#L45-L51

jcfr:
Other project integrating CMake:
* https://github.com/pyne/pyne/blob/develop/setup.py

jcfr:
@opadron Could you point out the part of the logic that should be improved along with an example of arguments that would be improperly parsed ? Thanks

opadron:
@msmolens did a good job improving the argument parsing logic.  I would consider this issue resolved.

jcfr:
Great. Thanks for the update.

opadron:
Scratch that, there are still some issues with the argument parsing.

scikit-build should recognize when command line options are given after the subcommand, and just pass them along.

For example,

`python setup.py sdist --help` causes the scikit-build logic to short-circuit straight to the main help output.

Instead, it should skip only the CMake-level build, and let the subcommand provide its help output.

opadron:
Scratch that, again!  It looks like the correct output is produced. I just didn't notice the subcommand help on the bottom.

Sorry for the noise!

PR #118, opened 2016-08-08
Improve argument parsing and --help output

### LINKED ISSUE #83: Improve python setup.py build --help output
- [ ] Document the different build type option
- [ ] Document how to set CMake variables
- [ ] Document how to set the CMake build system generator

msmolens:
Fixes #83 

Note that the scikit-build options are still treated as global options as opposed to build subcommand options. #57 hints that perhaps we want to push some options down into the subcommand.


jcfr:
Nicee :+1:

opadron:
This is great, @msmolens. On the subcommand matter, I don't think that work would be necessary.

PR #100, opened 2016-07-21
[WIP] Remove sample projects/Refactor CI driver

opadron:
Removes sample projects: "pen2-cython" and "tower-of-babel", in preparation for their move to their own projects repository.

Also, refactors more of the CI testing logic into driver scripts, in preparation for moving *it* to its own repository.

This PR is *super* WIP.  I expect the changes to break all the testing at first.

This PR should resolve or partially address #46, #63, #78 
This work should likely be merged with #97

PR #121, opened 2016-08-09
Update requirements

msmolens:


opadron:
LGTM, provided CI comes back green.

    Merging #121 into master will decrease coverage by 1.64%

Hmmm...how is that possible, I wonder?

opadron:
https://travis-ci.org/scikit-build/scikit-build/jobs/150977762#L5655
Ok, this is exactly what's happening to me over in #100.
Which is clearly due to something changing in travis.  Compare with a successful build less than two weeks ago.
https://travis-ci.org/scikit-build/scikit-build/jobs/148506325

I have to say, my trust in these free CI environments grows weaker every time they do something to break our builds.

DAE think having to hit a moving target somewhat defeats the purpose of a CI service?

jcfr:
> travis failure

for some reason, it fails on Linux + Python 3.3

>  moving target somewhat defeats the purpose of a CI service

Generally speaking I agree, that said we get the service for free. Our job to make test and code more robust

jcfr:
Here is the error:

```
----------------------------- Captured stderr call -----------------------------
CMake Warning:
  Manually-specified variables were not used by the project:
    SKBUILD
/usr/bin/ld: /opt/python/3.3.6/lib/libpython3.3m.a(longobject.o): relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a shared object; recompile with -fPIC
/opt/python/3.3.6/lib/libpython3.3m.a: could not read symbols: Bad value
collect2: ld returned 1 exit status
make[2]: *** [hello/_hello.so] Error 1
make[1]: *** [hello/CMakeFiles/_hello.dir/all] Error 2
make: *** [all] Error 2
  File "/home/travis/build/scikit-build/scikit-build/skbuild/setuptools_wrap.py", line 137, in setup
    cmkr.make(make_args)
  File "/home/travis/build/scikit-build/scikit-build/skbuild/cmaker.py", line 388, in make
    os.path.abspath(CMAKE_BUILD_DIR)))
```

opadron:
"Free" if you don't value your time.  I think, in the abstract, there's something to be said about how a hypothetical paid service that is more reliable might be preferable.

That being said, I'm willing to cut Travis some slack, since this is the first time I've seen *them* cause failing tests.  As we add more services, though, the chances that any one of them will cause problems on a given day will go up, and I think we should take care and make sure we don't hit a point of diminishing returns.

On the issue at hand, I see that @msmolens has already reported the problem to the Travis folks.
https://github.com/travis-ci/travis-ci/issues/6437

msmolens:
To allow us to proceed with development, we can allow a failure on the Travis CI/Linux/Python 3.3 for now: https://github.com/scikit-build/scikit-build/pull/122

opadron:
@msmolens Awesome.  Shall we close this, and continue with #122?

jcfr:
@msmolens Now #122 has been integrated, can you rebase this topic on master ? Thanks

opadron:
@msmolens Travis looks like it got hung up on Python 3.4?

opadron:
@msmolens Many thanks for this update.  I'll rebase for #100 and see if I can't get everything green, again.

PR #126, opened 2016-08-10
factor out custom command logic to mixins

opadron:
Removes code that was duplicated across multiple custom command classes and replaces it with a mixin solution.  Eliminates duplication and should also help with coverage numbers.

One technical challenge is that distutils/setuptools command classes are old-style classes, which don't work with mixins.  This PR addresses the challenge by introducing a small wrapper function that takes an old-style class and produces an equivalent new-style class.

Also, replaces superclass calls with proper uses of `super`.

### DIFF
--- a/skbuild/command/__init__.py
+++ b/skbuild/command/__init__.py
@@ -1,0 +1,23 @@
+
+from .. import cmaker
+
+
+class set_build_base_mixin(object):
+    def finalize_options(self, *args, **kwargs):
+        try:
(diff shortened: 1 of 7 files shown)


opadron:
Did everything just work the first time!?
Is this real life?

Wooo! :balloon: :dancers: :tada: :balloon: :tada: :dancers: :balloon:

thewtex:
:+1: Nice!

Issue #117, opened 2016-08-08
Packages must have at least one pure python module


thewtex:
Due to how the build extension modules are shipped as "package data", they are not shipped because:

  https://github.com/pypa/setuptools/blob/29c8385814335e91538d76b68578b6c0ab483c1a/setuptools/command/sdist.py#L147

One approach is to find some way to override distutils.dist.Distribution.has_pure_modules().

Another approach would check that there is at least one pure python module, and warn the user that this bug exists.

Another approach is to stop marking the generated C-extension as "package data", list them as the output of a custom *build_ext* command, and make sure they get added to the package manifest.

opadron:
@thewtex can you try the latest master, which should have #128, which according to @jcfr, should also fix this issue?

thewtex:
It does not look like #128 will effect this. The extension binary, which we are trying to pass along as package_data, will not get packaged because of the conditional here:

  https://github.com/pypa/setuptools/blob/29c8385814335e91538d76b68578b6c0ab483c1a/setuptools/command/sdist.py#L147

has_pure_modules() for distutils in Python 3.5.2 is:

```
    def has_pure_modules(self):
        return len(self.packages or self.py_modules or []) > 0
```

There does not appear to be any changes in #128 that change self.packages or self.py_modules.

opadron:
Ok, now I better understand what's happening.  If I'm not mistaken, you run into this issue when your setup call defines no modules or packages, and instead only has extension modules built in the cmake project?

If so, then this is actually intended, in the sense that I knew of this limitation when implementing.  The workaround I'd recommend is having a wrapping package or module that just imports all the symbols from the extension module.  Or, you can just provide an additional dummy package or module that does nothing.

I agree that this limitation is something we should correct, but I don't think it makes sense to even talk about how until we get significant progress on #124.  Work on that issue will impact so many aspects of how skbuild works, that this problem could very well just disappear, for example.

jcfr:
Closing. This is fixed by https://github.com/scikit-build/scikit-build/commit/ee10934d38f688687848c5ed347b9f4203b1e5bb

PR #128, opened 2016-08-29
setuptools_wrap: Force binary wheel generation. Fixes #106

### LINKED ISSUE #106: Generated wheels do not have a platform specfied
This is evident in the wheel file name (``none-any`) and in the METADATA file *Platform* entry, which is currently `UNKNOWN`.

jcfr:
Hack adapted from https://github.com/espdev/ITKPythonInstaller


jcfr:
This should also fix #117

PR #130, opened 2016-08-31
cmaker: Add support for CMake manifest file associated with components

jcfr:


opadron:
LGTM.  Thanks for the fix.

PR #131, opened 2016-08-31
Implement source distributions. fixes #84

### LINKED ISSUE #84: Fix python setup.py build sdist
Currently this creates a tarbal with the contents of _skbuild/cmake-install.
Instead, it should contain the project source code so that the sdist can be
downloaded and a platform-specific binary wheel can be created from it.

opadron:
Implements the `sdist` command for creating source distributions.
Uses the `sdist` command class from *distutils* (not setuptools).

Basically, it works by just jumping ahead of distutils' `sdist` logic and creating a MANIFEST file if the project does not provide its own.

This will only work within a git repo, since the implementation relies on the output of `git ls-files` to determine what files to include in the source distribution.

This implementation makes for some undesirable quirks, such as the output from `python setup.py sdist --help` being incorrect, but should suffice for the main source distribution needs for projects.

### DIFF
--- a/skbuild/command/egg_info.py
+++ b/skbuild/command/egg_info.py
@@ -37,7 +37,7 @@ class egg_info(set_build_base_mixin, new_style(_egg_info)):
 
         if do_generate:
             try:
-                with open('MANIFEST', 'w') as file:
+                with open('MANIFEST', 'wb') as file:
                     file.write(
                         subprocess.check_output(
                             ['git', 'ls-tree', '--name-only', '-r', 'HEAD'])
--- a/skbuild/command/egg_info.py
+++ b/skbuild/command/egg_info.py
@@ -45,9 +45,10 @@ class egg_info(set_build_base_mixin, new_style(_egg_info)):
             except subprocess.CalledProcessError:
                 sys.stderr.write(
                     '\n\n'
-                    'scikit-build tried to generate a MANIFEST '
-                    'file, but could not because the files to include could '
-                    'not be inferred.\n\n'
+                    'Since scikit-build could not find MANIFEST.in or '
+                    'MANIFEST, it tried to generate a MANIFEST file '
+                    'automatically, but could not because it could not '
+                    'determine which source files to include.\n\n'
                     'The command used was "git ls-tree --name-only -r HEAD"\n'
                     '\n\n'
                 )
--- a/skbuild/command/egg_info.py
+++ b/skbuild/command/egg_info.py
@@ -1,12 +1,63 @@
 
 import os
+import os.path
+import subprocess
+import sys
 
 from setuptools.command.egg_info import egg_info as _egg_info
 
 from . import new_style, set_build_base_mixin
+from ..cmaker import SKBUILD_DIR
 
+SKBUILD_MARKER_FILE = os.path.join(SKBUILD_DIR, "_skbuild_MANIFEST")
 
 class egg_info(set_build_base_mixin, new_style(_egg_info)):
+    def run(self):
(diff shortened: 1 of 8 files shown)


opadron:
argh! Forgot to test style...

skbuild/command/egg_info.py
```diff
@@ -1,12 +1,64 @@
...
+            try:
+                with open('MANIFEST', 'w') as file:
+                    file.write(
+                        subprocess.check_output(
+                            ['git', 'ls-tree', '--name-only', '-r', 'HEAD'])
+                    )
+            except subprocess.CalledProcessError:
+                sys.stderr.write(
+                    '\n\n'
+                    'scikit-build tried to generate a MANIFEST '
```
jcfr:
```
Since scikit-build could not find MANIFEST.in or MANIFEST, it tried to generate one automatically, but ...
```

jcfr: That would give a little more context

skbuild/command/egg_info.py
```diff
@@ -45,9 +45,10 @@ def run(self):
...
                 sys.stderr.write(
                     '\n\n'
-                    'scikit-build tried to generate a MANIFEST '
-                    'file, but could not because the files to include could '
-                    'not be inferred.\n\n'
+                    'Since scikit-build could not find MANIFEST.in or '
+                    'MANIFEST, it tried to generate a MANIFEST file '
+                    'automatically, but could not because it could not '
+                    'determine which source files to include.\n\n'
                     'The command used was "git ls-tree --name-only -r HEAD"\n'
```
jcfr: Please, also output source directory in which one the command was executed

opadron: It should be whatever the current directory is when the setup command is ran.  Do you think it's really necessary?

jcfr: Fair enough, you can ignore my comment

opadron:
https://ci.appveyor.com/project/scikit-build/scikit-build/build/0.0.1.99/job/4rrtskile5d660gb#L720
Looks like appveyor pulls in the working tree without keeping git's internal repo information.

I think we can use the following procedure to restore it:

```
git init
git remote add origin git://github.com/%APPVEYOR_REPO_NAME%
git fetch --all --prune
git reset --hard %APPVEYOR_REPO_COMMIT%
```

Does this look ok to everyone else?

skbuild/command/egg_info.py
jcfr: See https://ci.appveyor.com/project/scikit-build/scikit-build/build/0.0.1.99/job/4rrtskile5d660gb#L395

jcfr:
> appveyor pulls in the working tree without keeping git's internal repo information

Stange .. By default, it looks like AppVeyor clone repo with all its history 

https://www.appveyor.com/docs/how-to/repository-shallow-clone/#setting-depth-of-git-clone-command

> Does this look ok to everyone else?

Otherwise looks good

jcfr:
And reading here: https://www.appveyor.com/docs/how-to/git-push/

`Note: AppVeyor checks out only the last commit and not the entire branch. `

opadron:
@jcfr yeah, I just noticed we have shallow_clone set to true in the appveyor config.

I'll let this last CI complete just for science.  And then I'll try going back and just changing the config setting.

opadron:
@jcfr any idea why the codecov/patch check takes so long?

jcfr:
> any idea why the codecov/patch check takes so long?

Not sure why. Since CI turned green, I am integrating anyway.

PR #147, opened 2016-09-03
tests: Add test to check if bogus/unknown platform throw an exception

jcfr:


jcfr:
Integrated in https://github.com/scikit-build/scikit-build/commit/81829312860e63ab32682b97a0619dfb9bf7a32d

PR #156, opened 2016-09-08
skbuild: Do not raise error if running clean command twice

jcfr:
This commit prevents this error from being reported if clean command
is executed twice:

```
running clean
removing '_skbuild/cmake-install'
error: [Errno 2] No such file or directory: '_skbuild/cmake-install'
```


jcfr:
Superseded by #157

PR #158, opened 2016-09-08
skbuild/CMakePlatform: Remove unneeded subprocess call exception handling

jcfr:
Considering that:

* Check for invalid generator will not raise an exception, instead exit
  code of CMake indicates if there was an issue.

* Test for a valid installation of CMake is already done in the CMaker
  constructor by invoking "cmake --version" and catching OSError and
  CalledProcessError

This commit removes the extra try/catch.


jcfr:
Cc: @opadron

PR #159, opened 2016-09-09
tests: Check that the content of tar.gz sdist is correct

jcfr:


opadron:
One (slightly unrelated) thing I did forget to mention: I think on windows the source distribution is a Zip archive?  If so, you need to adjust the test slightly.

jcfr:
Very nice. 

Now we explicitly invoke the `egg_info` command, could we create our own command named `skbuild_collect_source` (or similar name) and invoke it instead ?  

Or do we have to derive from `egg_info` ?

opadron:
That is a possibility.  A similar idea I was thinking of was to see if somehow, we can use command classes to modify the arguments to `setup()` in a sort-of pipeline fashion.

If so, we can move a lot of our hackish cmake logic to command classes, and significantly simplify our command line argument handling logic.

tests/test_hello.py
```diff
@@ -81,24 +81,35 @@ def test_hello_sdist():
...
+        expected_content = [
+            'hello-1.2.3/CMakeLists.txt',
+            'hello-1.2.3/hello/_hello.cxx',
+            'hello-1.2.3/hello/CMakeLists.txt',
+            'hello-1.2.3/hello/__init__.py',
+            'hello-1.2.3/hello/__main__.py',
+            'hello-1.2.3/setup.py',
+            'hello-1.2.3/PKG-INFO'
+        ]
         member_list = ZipFile('dist/hello-1.2.3.zip').namelist()
```
jcfr: A little simpler:

```
+    expected_content = [
+        'hello-1.2.3/CMakeLists.txt',
+        'hello-1.2.3/hello/_hello.cxx',
+        'hello-1.2.3/hello/CMakeLists.txt',
+        'hello-1.2.3/hello/__init__.py',
+        'hello-1.2.3/hello/__main__.py',
+        'hello-1.2.3/setup.py',
+        'hello-1.2.3/PKG-INFO'
+    ]

+    member_list = None
+    expected_content = None
     if sdists_tar:
+        expected_content.extend([
+            'hello-1.2.3',
+            'hello-1.2.3/hello'
+        ])
         member_list = tarfile.open('dist/hello-1.2.3.tar.gz').getnames()
+
     elif sdists_zip:
        member_list = ZipFile('dist/hello-1.2.3.zip').namelist()
```

opadron: Once codecov passes, feel free to add it in.  The same simplification can be done in `test_hello_cython.py` and `test_hello_pure.py`.

jcfr: Roger that

opadron:
@jcfr viola!

jcfr:
Thanks.

Integrated in https://github.com/scikit-build/scikit-build/commit/3470742846e3398c8850f69a64680cbc709b65ca

PR #160, opened 2016-09-09
Run setup test in dedicated tmpdir

jcfr:
tests: Run each setup tests in a dedicated temp directory

This commit updates `project_setup_py_test` so that it automatically:
* creates a temporary directory
* copies the test project into it
* initializes associated git repository with one commit

It also remove the need for specifying `clear_cache` parameter. If the
project directory needs to be shared, the function decorated using
`project_setup_py_test` will return the tuple `(tmp_dir, decorated_func_result)`.

Then `tmp_dir` can be passed as a parameter to an other call to `project_setup_py_test`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,6 +51,4 @@ docs/_build
 
 # build output (testing)
 skbuild/_cmake_test_compile/*
-tests/samples/*/_cmake_test_compile/*
-tests/samples/*/MANIFEST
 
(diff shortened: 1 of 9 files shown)


jcfr:
@opadron Let me know what you think of this.

jcfr:
Integrated in https://github.com/scikit-build/scikit-build/commit/d1664ee21358a44513036f2624faeb695af283cc

PR #161, opened 2016-09-10
utils: Simplify code introducing ContextDecorator. Fixes #154

### LINKED ISSUE #154: Use ContextDecorator instead of utils.decorator
context manager (in `utils/__init__.py`) and decorator (in `utils/decorator.py`) could be re-factored into a single definition by using `ContextDecorator` 

See:
*  https://coderwall.com/p/0lk6jg/python-decorators-vs-context-managers-have-your-cake-and-eat-it
* http://contextlib2.readthedocs.io/en/latest/
* http://stackoverflow.com/questions/9213600/function-acting-as-both-decorator-and-context-manager-in-python#9213668

jcfr:
This commit introduces a base class allowing to create context manager
that can also be used a decorator.

Adapted from https://coderwall.com/p/0lk6jg/python-decorators-vs-context-managers-have-your-cake-and-eat-it


jcfr:
Integrated in https://github.com/scikit-build/scikit-build/commit/11616ad0bb97d481fe29390dd2d8cc7d20aa97a2

PR #171, opened 2016-09-12
Support pure python package and package built by cmake

jcfr:
setuptools_wrap: Add support for pure python packages and packages built by CMake

Prior to this commit, install rules for all packages were required
to be added to the CMake project. This commit changes this by allowing
mixing of pure python package along with package built by CMake.

To illustrate this, the "hello" sample project has been updated to include
a pure python package named 'bonjour' that doesn't have any install
rules in any of the CMakeLists.txt


jcfr:
Integrated in https://github.com/scikit-build/scikit-build/commit/74f72270b4a0469813cdb775188518ac6dbcf88d

PR #178, opened 2016-09-16
tests: Use "to/the/base" instead of "base" in test_setup_inputs

jcfr:


jcfr:
Closing. Change integrated in https://github.com/scikit-build/scikit-build/commit/7a9fb3e3a4edaeeb8c3685160bc017afa0149bff

PR #180, opened 2016-09-19
Add find numpy module

jcfr:


jcfr:
Cc: @thewtex

jcfr:
It turns out that `NumPy_VERSION` is not a cache variable. 

The version is always retrieved by reading the value of `numpy.__version__`. This means that the NumPy version is not fully described by `PYTHON_INCLUDE_DIR`.

Without a definite use case at hand, I think the current solution is sufficient

jcfr:
Integrated in https://github.com/scikit-build/scikit-build/commit/9d882f98727a934af3129560a58f154867b11110

PR #190, opened 2016-09-25
Fix targetLinkLibrariesWithDynamicLookup

jcfr:
This commit has two purposes:

(1) it fixes logic checking for cross-compilation (the regression
was introduced by scikit-build/scikit-build#51 and scikit-build/scikit-build#47

(2) It configure the text project setting CMAKE_ENABLE_EXPORTS to ON. Doing
so ensure the executable compiled in the test exports symbols (if supported
by the underlying platform).
See https://cmake.org/cmake/help/v3.6/prop_tgt/ENABLE_EXPORTS.html?highlight=enable_export

### DIFF
--- a/skbuild/resources/cmake/targetLinkLibrariesWithDynamicLookup.cmake
+++ b/skbuild/resources/cmake/targetLinkLibrariesWithDynamicLookup.cmake
@@ -327,6 +327,7 @@ function(_test_weak_link_project
                 "${project_name}"
                 CMAKE_FLAGS
                   "-DCMAKE_SHARED_LINKER_FLAGS='${CMAKE_SHARED_LINKER_FLAGS}'"
+                  "-DCMAKE_ENABLE_EXPORTS=ON"
                   ${_rpath_arg}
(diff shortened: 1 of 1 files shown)


jcfr:
Cc: @opadron @thewtex 

This commit fixes the module `targetLinkLibrariesWithDynamicLookup`, the existing implementation
was systematically linking against the given libraries.

I was able to identify the issues by working on improving and back-porting changes to `weak-linking-demo`
project. See https://github.com/opadron/weak-linking-demo/pull/6

jcfr:
Ideally, we should transfer the project https://github.com/opadron/weak-linking-demo to scikit-build organization and rename it to `cmake-module-targetLinkLibrariesWithDynamicLookup`.

That way, we would have a ground to develop and **test** using CI

thewtex:
LGTM!

jcfr:
Thanks for the review. 

Before calling it a victory, I will also add a test and address https://github.com/jcfr/SimpleITKPythonPackage/issues/5

PR #202, opened 2016-10-26
Tweak doc

jcfr:


jcfr:
Cc: @thewtex

docs/extension_build_system.rst
```diff
@@ -0,0 +1,31 @@
+======================
+Extension Build System
+======================
+
+
+Introduction
+------------
+
+By default, scikit-build looks in the project top-level directory for a
+file named ``CMakeLists.txt``. It will then invoke CMake executable specifying
```
thewtex: invoke CMake executable -> invoke the `cmake` executable

docs/extension_build_system.rst
```diff
@@ -0,0 +1,31 @@
...
+a generator matching the python being used.
+
+Indeed, each python version is associated with an official compiler. By default,
```
thewtex: python -> CPython

docs/extension_build_system.rst
```diff
@@ -0,0 +1,31 @@
...
+scikit-build will automatically select the generator matching the official
```
thewtex: generator -> compiler, associated C runtime, and build flags

I think these or similar are more specific. Also, hopefully in the future we can use the Ninja generator.

docs/extension_build_system.rst
```diff
@@ -0,0 +1,31 @@
...
+recommendations:
+
+.. note:: *Add table here + link to official doc*
+
+
+
+How to test if scikit-build is used ?
```
thewtex: is used -> is driving the compilation

jcfr:
Thanks again for the review, documentation updated based on your suggestion and PR integrated.

PR #224, opened 2016-12-09
Fallback to use ms compiler for python27

jcfr:
This topic is still **WORK IN PROGRESS**, the intent of this PR it to validate the work done so far.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,19 @@ New features
 
 * Add support for `development mode <https://packaging.python.org/distributing/#working-in-development-mode>`_. (:issue:`187`).
 
+* If available, uses `Ninja <https://cmake.org/cmake/help/v3.7/generator/Ninja.html>`_
+  build system generator on all platforms. An advantages is that ninja automatically
+  parallelizes the build based on the number of CPUs. See :doc:`/generators`.
+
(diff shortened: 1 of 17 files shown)


jcfr:
This topic is getting ready for final review and integration.

It adds support for ``Ninja``, ``NMake Makefiles`` and ``NMake Makefiles JOM`` command line generators
 with automatic settings of the `Visual Studio` environment expected by the version of python being used.

This is made possible by internally using the function ``query_vcvarsall`` from the ``distutils.msvc9compiler`` (or ``distutils._msvccompiler`` when visual studio ``>= 2015`` is used). 

Note that since the modules are systematically patched using ``setuptools.monkey.patch_for_msvc_specialized_compiler()``, skbuild will also properly work when [Microsoft Visual C++ Compiler for Python 2.7](http://aka.ms/vcpython27) is  available.

Notes:
* Documentation will be updated shortly
* Specifying `-G Ninja` will always look for the version of visual studio associated with the version of python being used.
* IOW, it is not yet possible to specify `-G Ninja` and select a version of visual studio **different from the one officially supported by the python being used**. This could be done by teaching scikit-build a new argument allowing to choose a *build environment" and a platform (e.g `python setup.py build -G "Ninja" -B "Visual Studio 2015" -A x64`).

Cc: @thewtex

PR #234, opened 2016-12-19
Address CMP0054 warning with "unused" string/variable

blowekamp:
Use standard regex MATCH approach to avoid a string being confused
with a variable. This patch addressed the following CMake warning:

```
CMake Warning (dev) at
CMake/sitkTargetLinkLibrariesWithDynamicLookup.cmake:466 (if):
  Policy CMP0054 is not set: Only interpret if() arguments as
  variables or
  keywords when unquoted.  Run "cmake --help-policy CMP0054" for
  policy
  details.  Use the cmake_policy command to set the policy and
  suppress this
  warning.

  Quoted variables like "unused" will no longer be dereferenced when
  the
  policy is set to NEW.  Since the policy is not set the OLD behavior
  will be
  used.
```

### DIFF
--- a/skbuild/resources/cmake/targetLinkLibrariesWithDynamicLookup.cmake
+++ b/skbuild/resources/cmake/targetLinkLibrariesWithDynamicLookup.cmake
@@ -463,7 +463,7 @@ function(check_dynamic_lookup)
     ${link_flags_var}
     )
   set(${has_dynamic_lookup_var} ${${has_dynamic_lookup_var}} PARENT_SCOPE)
-  if(NOT link_flags_var STREQUAL "unused")
+  if(NOT "x${link_flags_var}x" STREQUAL "xunusedx")
     set(${link_flags_var} ${${link_flags_var}} PARENT_SCOPE)
   endif()
 endfunction()


jcfr:

Makes sense, this will allow to work nicely with older version of CMake

skbuild/resources/cmake/targetLinkLibrariesWithDynamicLookup.cmake
```diff
@@ -463,7 +463,7 @@ function(check_dynamic_lookup)
     ${link_flags_var}
     )
   set(${has_dynamic_lookup_var} ${${has_dynamic_lookup_var}} PARENT_SCOPE)
-  if(NOT link_flags_var STREQUAL "unused")
+  if(NOT "x${link_flags_var}x" MATCHES "^xunusedx$")
```
jcfr: If that works with you, I will plan on using a simpler form:
```
if(NOT "x${link_flags_var}" STREQUAL "xunused")
```

blowekamp:
IMHO, using MATCH with the addition of the special regex character '^' and '$' virtually ensures that there is not a variable with that name. Thereby ensuring very good protection from the CMP issue. What is the disadvantage?

jcfr:
Since we are simply comparing string, I think the form with `STREQUAL` is more appropriate.

A recent commit from CMake core dev also seems to indicate using `STREQUAL` is sufficient. See https://github.com/Kitware/CMake/commit/be73273952974e15ed12cf41180162bbbda70da3

Also worth noting the `STREQUAL` form is what is used in the CMake code base to protect simple string comparison, I thought to make things consistent. See below.

All of that said, if using `MATCHES`  protects from issues not covered by `STREQUAL`, we should look into also updating CMake code base.

```
cd /path/to/source/cmake
$ ack --cmake "x\\$\{.+STREQUAL" -l | wc -l
67
$ ack --cmake "x\\$\{.+MATCHES" -l | wc -l
13

```

blowekamp:
Thanks for the information on how CMake does it. I wonder where I got that this approach was a "best practice".

jcfr:
Thanks. Integrated in https://github.com/scikit-build/scikit-build/commit/b53a7915c16672fffa39186c4fde2e6f19f63086

Issue #240, opened 2017-01-25
targetLinkLibrariesWithDynamicLookup does not cache link_flags


blowekamp:
When run on OS X, the required link flags are lost sometime when CMake is run multiple times.

jcfr:
Closing. Fixed by https://github.com/scikit-build/scikit-build/commit/2c0e574f1f619d366ea930f46614f07a48a0033d

PR #241, opened 2017-01-25
Correct DYNAMIC_LOOKUP_* cache type to STRING

blowekamp:
When CMake ran multiple time, the value would change from the correct
string to a boolean value. This caused failure of the expected dynamic
lookup behavior.

This patch addresses #240


jcfr:
Thanks for the PR :+1:

PR #254, opened 2017-04-12
skbuild.plaform_specifics: pass CMake args to the generator check

thewtex:
The CMake command line arguments should be passed generator check cmake runs.
This ensures that builds can be successful if they require compiler flags that
were passed or if they require CMake variables to be set like
CMAKE_MAKE_PROGRAM.


jcfr:
Closing. Superseded by #256

PR #260, opened 2017-06-11
Fix logical error of generating MANIFEST.

seanlis:


jcfr:
@seanlis Thanks for the patch :+1: 

Would you have the steps allowing to reproduce the problem ? Ideally, I would like to add a regression test capturing the problem fixed by our contribution.

Thanks a lot.
Jc

seanlis:
1. There is a MANIFEST.in in the project.
2. There is no MANIFEST file.
3. Run 'python setup.py sdist'
4. A MANIFEST file would be generated by skbuild from the output of
        `git ls-tree --name-only -r HEAD` instead `MANIFEST.in`.

jcfr:
@seanlis Thanks for the update. 

Just to let you that I will be travelling for the next week and I am not sure if I will  have time to follow up until I am back.

henryiii:
Can this one-word change patch be applied? It's a critical bug that makes it impossible to use a `MANIFEST.in`, since if `MANIFEST` exists, `MANIFEST.in` will not write to it.  I've manually made this patch to get the `MANIFEST.in` working before I found this PR.

It might be easier to work the logic through if you make the variable `do_not_generate` and then `not` it, since you're really checking for reasons not to try to generate the file.

jcfr:
Closing. This is superseded by #310

PR #273, opened 2018-01-24
Add findf2py

xoviat:

PR #276, opened 2018-02-09
Find f2py rebased

jcfr:
From https://github.com/scikit-build/scikit-build/pull/273#issuecomment-362862052:

So the idea here is actually to eliminate `numpy.distutils` entirely over time and replace it with scikit-build. We can't do that if we write `f2py -c` because `numpy.distutils` is still used. Instead we want to use `f2py` to generate the C files and then link them to the Python extension.

I'm also not happy with the way that Cython support was implemented in scikit-build. Ideally, you would be able to write the following and have scikit-build autogenerate the cython and f2py targets:

```
add_library(LIBRARY_NAME
    input.pyf.src
    input.pyx
) 
```

If you look at [what we're replacing](https://github.com/scipy/scipy/blob/5531da38cec429f3cf2f9aca0f698e2130562592/scipy/fftpack/setup.py), we essentially need to translate the `add_library` and `add_extension` function calls on a one-to-one basis.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,14 @@ details, see the commit logs at http://github.com/scikit-build/scikit-build
 Next Release
 ============
 
+New Features
+------------
+
+* Add CMake module :doc:`\cmake-modules/F2PY` useful to find the ``f2py`` executable for building Python
+  extensions with Fortran. Thanks to :user:`xoviat` for moving forward with the integration. Concept for the
+  module comes from the work of :user:`scopatz` done in `PyNE <https://github.com/pyne/pyne>`_ project.
+  See :issue:`273`.
+
 Bug fixes
 ---------
 
--- a/docs/cmake-modules.rst
+++ b/docs/cmake-modules.rst
@@ -12,6 +12,7 @@ CMake modules:
    cmake-modules/Cython
    cmake-modules/NumPy
    cmake-modules/PythonExtensions
+   cmake-modules/F2PY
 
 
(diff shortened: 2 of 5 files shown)


jcfr:
Before integrating, I would like to have a working example. You will find [here](https://github.com/scikit-build/scikit-build-sample-projects/tree/add-f2py-sample-projects/projects/hello-f2py#readme) two sample projects:
* `hello`: Approach where the module source is generated using `f2py` and then explicitly compiled into a module
* `hello_auto`: Approach where `f2py` generate the compiled module.

See https://github.com/scikit-build/scikit-build-sample-projects/tree/add-f2py-sample-projects/projects/hello-f2py#readme

There is some error with the `auto` case that I would like to address first.

It would also be nice to extend the `FindF2PY` module with some convenience function to interface with `f2py`.

xoviat:
> There are some errors with the auto case that I would like to address first.

I don't think we should address these errors. None of the compilation options will be passed because CMake won't be in charge of the compilation. 

> It would also be nice to extend the FindF2PY module with some convenience function to interface with f2py.

I agree. However, I do believe in incremental development, which means breaking this up into multiple pull requests rather than doing it all at once.

jcfr:
@xoviat This is ready for final review. 

(Since the module ended up being completely rewritten, I removed the license from pyne and acknowledged @scopatz  in the change log for the concept.)

jcfr:
I integrated the changes.

@xoviat Thanks again for your contribution 👍

PR #278, opened 2018-02-15
Add conv-template and from-template scripts

xoviat:
These scripts are required to generate *.src sources present in NumPy and SciPy.

cc @jcfr

### DIFF
--- a/skbuild/resources/cmake/FindNumPy.cmake
+++ b/skbuild/resources/cmake/FindNumPy.cmake
@@ -48,7 +48,7 @@ if(NOT NumPy_FOUND)
       ERROR_QUIET
       )
 
-    if(NOT NumPy_CONV_TEMPLATE_EXECUTABLE)
+    if(NOT NumPy_CONV_TEMPLATE_COMMAND)
       execute_process(COMMAND "${PYTHON_EXECUTABLE}"
         -c "from numpy.distutils import conv_template; print(conv_template.__file__)"
         OUTPUT_VARIABLE _numpy_conv_template_file
@@ -58,7 +58,7 @@ if(NOT NumPy_FOUND)
       set(NumPy_CONV_TEMPLATE_COMMAND "${PYTHON_EXECUTABLE}" "${_numpy_conv_template_file}")
     endif()
 
-    if(NOT NumPy_FROM_TEMPLATE_EXECUTABLE)
+    if(NOT NumPy_FROM_TEMPLATE_COMMAND)
       execute_process(COMMAND "${PYTHON_EXECUTABLE}"
         -c "from numpy.distutils import from_template; print(from_template.__file__)"
         OUTPUT_VARIABLE _numpy_from_template_file
--- a/skbuild/resources/cmake/FindNumPy.cmake
+++ b/skbuild/resources/cmake/FindNumPy.cmake
@@ -10,9 +10,9 @@
 #   The include directories needed to use NumpPy.
 # ``NumPy_VERSION``
 #   The version of NumPy found.
-# ``NumPy_CONV_TEMPLATE``
+# ``NumPy_CONV_TEMPLATE_EXECUTABLE``
 #   The command-line arguments required to execute the conv-template script
-# ``NumPy_FROM_TEMPLATE``
+# ``NumPy_FROM_TEMPLATE_EXECUTABLE``
 #   The command-line arguments required to execute the from-template script
 #
 # The module will also explicitly define one cache variable:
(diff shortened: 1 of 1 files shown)


jcfr:
Thanks for the patch 👍 

You will find few comments posted below.

skbuild/resources/cmake/FindNumPy.cmake
```diff
@@ -10,7 +10,10 @@
 #   The include directories needed to use NumpPy.
 # ``NumPy_VERSION``
 #   The version of NumPy found.
-#
+# ``NumPy_CONV_TEMPLATE``
+#   The command-line arguments required to execute the conv-template script
+# ``NumPy_FROM_TEMPLATE``
+#   The command-line arguments required to execute the from-template script
```
jcfr: I suggest to name the variables `NumPy_CONV_TEMPLATE_EXECUTABLE` and  `NumPy_FROM_TEMPLATE_EXECUTABLE`.

Since the variable is set to the path to the executable, should the description be updated to something like this:

```
Path to the conv-template script
```

xoviat: Well it's not necessarily the executable. It's supposed to be one or two arguments that you add to the beginning of the command line.

jcfr: Since `conv-template` and `from-template` are python scripts, using `find_program` will not work.

Instead I suggest the following approach:

```
if(NOT DEFINED NumPy_CONV_TEMPLATE_SCRIPT)
  execute_process(COMMAND "${PYTHON_EXECUTABLE}"
       -c "from numpy.distutils import conv_template; print(conv_template.__file__)"
    OUTPUT_VARIABLE _numpy_conv_template_file
    OUTPUT_STRIP_TRAILING_WHITESPACE
    ERROR_QUIET
  )
  set(NumPy_CONV_TEMPLATE_SCRIPT "${_numpy_conv_template_file}" CACHE FILEPATH "path to conv-template script")
endif()

...
```

jcfr:
This approach would give the most flexibility:

```
if(NOT DEFINED NumPy_CONV_TEMPLATE_SCRIPT)
  execute_process(COMMAND "${PYTHON_EXECUTABLE}"
       -c "from numpy.distutils import conv_template; print(conv_template.__file__)"
    OUTPUT_VARIABLE _numpy_conv_template_file
    OUTPUT_STRIP_TRAILING_WHITESPACE
    ERROR_QUIET
  )
  set(NumPy_CONV_TEMPLATE_SCRIPT "${_numpy_conv_template_file}" CACHE FILEPATH "path to conv-template script")
endif()
```

On the other hand, if having both the python interpreter and the script packed into one variable is important, the variable could be named `NumPy_CONV_TEMPLATE_COMMAND`

skbuild/resources/cmake/FindNumPy.cmake
xoviat: These scripts will be present in future releases of NumPy though (see [here](https://github.com/numpy/numpy/blob/5c5a215fa1101479ae9b8d127be32679c9f3f105/setup.py#L366-L367)). The `execute_process` is for backward compatibility only.

xoviat:
These scripts will be present in future releases of NumPy though (see [here](https://github.com/numpy/numpy/blob/5c5a215fa1101479ae9b8d127be32679c9f3f105/setup.py#L366-L367)). The `execute_process` is for backward compatibility only.

jcfr:
Gotcha. 

Since this is for backward compatibility .... naming the variable with `_EXECUTABLE` prefix make sense.  (Sorry for the back and forth ...)

Let's also make sure to set the variable into the cache.

Here is what I suggest:

```
  find_program(NumPy_CONV_TEMPLATE_EXECUTABLE NAMES conv-template)
  find_program(NumPy_FROM_TEMPLATE_EXECUTABLE NAMES from-template)

  if(PYTHON_EXECUTABLE)
   
    [...]

    # XXX To support NumPy < v0.15.0 where conv-template is not declared as an entry point, we
    # emulate the behavior of a standalone executable using the path the the python interpreter and the 
    # associated conv-template script.
    if(NOT NumPy_CONV_TEMPLATE_EXECUTABLE)
      execute_process(COMMAND "${PYTHON_EXECUTABLE}"
        -c "from numpy.distutils import conv_template; print(conv_template.__file__)"
        OUTPUT_VARIABLE _numpy_conv_template_file
        OUTPUT_STRIP_TRAILING_WHITESPACE
        ERROR_QUIET
        )
      set(NumPy_CONV_TEMPLATE_EXECUTABLE "${PYTHON_EXECUTABLE}" "${_numpy_conv_template_file}" CACHE STRING "Command executing conv-template program")
    endif()

    # XXX To support NumPy < v0.15.0 where from-template is not declared as an entry point, we
    # emulate the behavior of a standalone executable using the path the the python interpreter and the 
    # associated from-template script.
    if(NOT NumPy_FROM_TEMPLATE_EXECUTABLE)
      execute_process(COMMAND "${PYTHON_EXECUTABLE}"
        -c "from numpy.distutils import from_template; print(from_template.__file__)"
        OUTPUT_VARIABLE _numpy_from_template_file
        OUTPUT_STRIP_TRAILING_WHITESPACE
        ERROR_QUIET
        )
      set(NumPy_FROM_TEMPLATE_EXECUTABLE "${PYTHON_EXECUTABLE}" "${_numpy_from_template_file}" CACHE STRING "Command executing from-template program")
    endif()
  endif()
endif()
```

xoviat:
Does that look okay?

jcfr:
Closing. Integrated in https://github.com/scikit-build/scikit-build/commit/7abf44c259d7b44781b214d7e8a9f75e77d680cb

Issue #274, opened 2018-01-28
pip install fails when package_dir = {'':'src'}


benjaminjack:
I have a package structure that looks like this:

```
├─ src
│  └─ packagename
│     ├─ __init__.py
│     └─ ...
├─ tests
│  └─ ...
├─ CMakeLists.txt
└─ setup.py
```

My setup.py script looks like this:

```
setup(
    ...
    packages=find_packages('src'),
    package_dir={'': 'src'},
    ...
)
```

Running `pip install .` and `pip install -e .` fails with an error saying that directory 'package_name' cannot be found. Running the same commands with setuptools setup works fine.

jcfr:
@benjaminjack  Thanks for the report :+1: 

This should be fixed by https://github.com/scikit-build/scikit-build/pull/296

Once integrated, you should be able to also update `benjaminjack/python_cpp_example` project by integrating  https://github.com/benjaminjack/python_cpp_example/pull/1

jcfr:
Closing. Fixed integrated in https://github.com/scikit-build/scikit-build/commit/3f5cfc659e4270f50a55d6e0b0bee0561e9c520b

anthrotype:
@jcfr thanks for fixing this! I also use the same package layout.
I see this is fixed on master, but not yet on the current version which is on PyPI.
Do you have an estimate when the next version will be released and uploaded to PyPI?
I know I can pip install directly from the github repo for the time being, but still would be nice to have it on PyPI as well.
Thank you

jcfr:
Good news. A new release will be pushed any day !

PR #296, opened 2018-02-21
setuptools_wrap: Add support for default package_dir

jcfr:
This commit supports specifying a default directory for all packages
that do not have one already explicitly associated with them.

See #274


jcfr:
A related regression was addressed in https://github.com/scikit-build/scikit-build/pull/319

PR #298, opened 2018-02-21
UseCython: Add fixes and new features

xoviat:
- Add ability to read C or CXX from source properties
- Add DEPENDS to Cython command
- Use --fast-fail to improve build output
- Improve handling of generated Cython files


xoviat:
@jcfr Can you guess the reason for the failures? I cannot reproduce locally.

xoviat:
I'm marking this as WIP for now so that I can try for a better solution.

skbuild/resources/cmake/UseCython.cmake
```diff
@@ -205,15 +205,22 @@ function(add_cython_target _name)
 
   set(comment "Generating ${_output_syntax} source ${generated_file_relative}")
   set(cython_include_directories "")
-  set(pxd_dependencies "")
+  set(pxd_dependencies "${args_DEPENDS}")
   set(c_header_dependencies "")
 
+  # Filter pxd_dependencies so that it only contains pxd files
+  list(FILTER pxd_dependencies INCLUDE REGEX .pxd$)
```
xoviat: Period needs to be escaped.

skbuild/resources/cmake/UseCython.cmake
```diff
@@ -286,8 +298,12 @@ function(add_cython_target _name)
       list(REMOVE_DUPLICATES include_dependencies)
 
       # Add modules to the files to check, if appropriate.
-      foreach(module ${module_dependencies})
+      foreach(module IN LISTS module_dependencies)
         unset(pxd_location CACHE)
+        string(REPLACE "." "/" module ${module})
+        # TODO: Handle relative import paths, as well as establishing
+        #       The project root directory
+
```
xoviat: I'm not exactly sure how to handle this. The module appears designed to find `x.y.z.pxd` files, but in reality modern Cython uses `x/y/z.pxd` files. What if we need to find the pxd from `cimport x.y.z`? Are we to assume that it could be in `${CMAKE_SOURCE_DIR}/x/y/z.pxd`? What justifies that assumption, especially with gh-296?

Thoughts @jcfr?

Issue #295, opened 2018-02-21
Managing symbol visibility


ghost:
On windows we need to do this:

```
#  set_target_properties(${_name} PROPERTIES LINK_FLAGS 
#      "/EXPORT:PyInit_${_name}"
#  )
```

On linux, we need to use the linker script (`--version-script`) to hide symbols:
```
{
  global:
    PyInit_${_name};
  local: *;
}
```

jcfr:
Closing. Changes integrated in 8b15eff

PR #299, opened 2018-02-22
FindPythonExtensions: set symbol visibility correclty

xoviat:
In addition to aligning the behavior of scikit-build with distutils, this PR attempts a generic fix to issues such as scipy/scipy#8451.

Closes gh-295.


skbuild/resources/cmake/FindPythonExtensions.cmake
```diff
@@ -317,6 +317,36 @@ if(NOT DEFINED PYTHON_EXTENSION_MODULE_SUFFIX)
...
+  get_target_property(LANG ${_target} LINKER_LANGUAGE)
+  set(_linker_id "${CMAKE_${LANG}_COMPILER_ID}")
+
+  if(MSVC)
+    set_target_properties(${_target} PROPERTIES LINK_FLAGS 
+        "/EXPORT:PyInit_${_target}"
+    )
+  elseif("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU" AND
+         "${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" AND
+         "${CMAKE_Fortran_COMPILER_ID}" STREQUAL "GNU")
```
xoviat: What I really want to say is that "this extensions will be linked with ld." If there's a better way to do that, please say so.

skbuild/resources/cmake/FindPythonExtensions.cmake
```diff
@@ -317,6 +317,31 @@ if(NOT DEFINED PYTHON_EXTENSION_MODULE_SUFFIX)
   set(PYTHON_EXTENSION_MODULE_SUFFIX "${_item}")
 endif()
 
+function(_set_python_extension_symbol_visibility _target)
```
jcfr: To support both python2 and python3, I suggest to have something like this:

```
function(_set_python_extension_symbol_visibility _target)
[...]
  if(PYTHON_VERSION_MAJOR VERSION_GREATER 2)
    set(module_init_function_prefix "PyInit_")
  else()
    set(module_init_function_prefix "init")
  endif()

  if("${CMAKE_C_COMPILER_ID}" STREQUAL "MSVC")
    set_target_properties(${_target} PROPERTIES LINK_FLAGS 
        "/EXPORT:${module_init_function_prefix}${_target}"
    )
  elseif("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU")
    set(_script_path
      ${CMAKE_CURRENT_BINARY_DIR}/_skbuild/${_target}-version-script.map
    )
    file(WRITE
      ${_script_path}
      "
      {
        global:
          ${module_init_function_prefix}${_target};
        local: *;
      }
      "      
    )
    set_target_properties(${_target} PROPERTIES LINK_FLAGS 
        "-Wl,--version-script=${_script_path}"
    )
  endif()
endfunction()
```

jcfr: Also, instead of `_skbuild`, we should introduce the equivalent of `SKBUILD_DIR` (already available in python as `skbuild.constant.SKBUILD_DIR`) in CMake

jcfr: Since extension module includes `PyMODINIT_FUNC`, the associated symbol are already exported. For example, see [here](https://github.com/scikit-build/scikit-build/blob/3f5cfc659e4270f50a55d6e0b0bee0561e9c520b/tests/samples/hello-cpp/hello/_hello.cxx#L61)

On linux, using `readelf -s /path/to/module.so` also confirmed that the symbol is already global.

I suspect that visibility need to be set only for the cython generated modules.

xoviat: The purpose of the script on Linux to hide all of the other symbols. Can you confirm the script has the intended effect?

skbuild/resources/cmake/FindPythonExtensions.cmake
```diff
@@ -317,6 +317,37 @@ if(NOT DEFINED PYTHON_EXTENSION_MODULE_SUFFIX)
...
+      ${_script_path}
+      "
+      {
+        global:
+          ${_modinit_prefix}${_target};
+        local: *;
+      }
+      "      
+    )
+    set_target_properties(${_target} PROPERTIES LINK_FLAGS
```
jcfr: It turns out that this a compile flags and not a link flags. `COMPILE_FLAGS` should be used

xoviat: On windows, link flags are used (I think).

jcfr: I talked to fast ... in fact the following is happening:
* `_set_python_extension_symbol_visibility` set the linker flags as it should 
* but these are reset to `-Wl,--unresolved-symbols=ignore-all` when gnu is used. See function `target_link_libraries_with_dynamic_lookup` called lower in this file. 

I will try to call the function `_set_python_extension_symbol_visibility` afterward

jcfr: Now that the linker is flag is passed, here is the error:

```
[100%] Linking CXX shared module _hello.cpython-35m-x86_64-linux-gnu.so
cd /tmp/pytest-of-jcfr/pytest-jcfr/_skbuild/linux-x86_64-3.5/cmake-build/hello && /home/jcfr/Software/cmake-3.10.0-Linux-x86_64/bin/cmake -E cmake_link_script CMakeFiles/_hello.dir/link.txt --verbose=1
/usr/bin/c++ -fPIC -O3 -DNDEBUG -Wl,--version-script=/tmp/pytest-of-jcfr/pytest-jcfr/_skbuild/linux-x86_64-3.5/cmake-build/hello/_skbuild/_hello-version-script.map -shared  -o _hello.cpython-35m-x86_64-linux-gnu.so CMakeFiles/_hello.dir/_hello.cxx.o 
/usr/bin/ld:/tmp/pytest-of-jcfr/pytest-jcfr/_skbuild/linux-x86_64-3.5/cmake-build/hello/_skbuild/_hello-version-script.map:7: syntax error in VERSION script
collect2: error: ld returned 1 exit status
hello/CMakeFiles/_hello.dir/build.make:94: recipe for target 'hello/_hello.cpython-35m-x86_64-linux-gnu.so' failed
make[2]: *** [hello/_hello.cpython-35m-x86_64-linux-gnu.so] Error 1

```

xoviat: If that is the case, then it's probably best to append the version script to the `-Wl,--unresolved-symbols=ignore-all` argument.

xoviat:
@jcfr Are the issues addressed now?

jcfr:
With the last fix, it seems to have the intended effect:

### With version script

```
Symbol table '.dynsym' contains 14 entries:
   Num:    Value          Size Type    Bind   Vis      Ndx Name
     0: 0000000000000000     0 NOTYPE  LOCAL  DEFAULT  UND 
     1: 0000000000000708     0 SECTION LOCAL  DEFAULT    9 
     2: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _ITM_deregisterTMCloneTab
     3: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND __stack_chk_fail@GLIBC_2.4 (2)
     4: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND PyLong_FromLong
     5: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND __gmon_start__
     6: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND PyModule_Create2
     7: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND _Py_NoneStruct
     8: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND PyArg_ParseTuple
     9: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _Jv_RegisterClasses
    10: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _ITM_registerTMCloneTable
    11: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND PySys_WriteStdout
    12: 0000000000000000     0 FUNC    WEAK   DEFAULT  UND __cxa_finalize@GLIBC_2.2.5 (3)
    13: 0000000000000930    17 FUNC    GLOBAL DEFAULT   11 PyInit__hello
```

### Without version script

```
Symbol table '.dynsym' contains 19 entries:
   Num:    Value          Size Type    Bind   Vis      Ndx Name
     0: 0000000000000000     0 NOTYPE  LOCAL  DEFAULT  UND 
     1: 00000000000007c8     0 SECTION LOCAL  DEFAULT    9 
     2: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _ITM_deregisterTMCloneTab
     3: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND __stack_chk_fail@GLIBC_2.4 (2)
     4: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND PyLong_FromLong
     5: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND __gmon_start__
     6: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND PyModule_Create2
     7: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND _Py_NoneStruct
     8: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND PyArg_ParseTuple
     9: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _Jv_RegisterClasses
    10: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _ITM_registerTMCloneTable
    11: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND PySys_WriteStdout
    12: 0000000000000000     0 FUNC    WEAK   DEFAULT  UND __cxa_finalize@GLIBC_2.2.5 (3)
    13: 0000000000201160     0 NOTYPE  GLOBAL DEFAULT   22 _edata
    14: 0000000000201168     0 NOTYPE  GLOBAL DEFAULT   23 _end
    15: 00000000000009f0    17 FUNC    GLOBAL DEFAULT   11 PyInit__hello
    16: 0000000000201160     0 NOTYPE  GLOBAL DEFAULT   23 __bss_start
    17: 00000000000007c8     0 FUNC    GLOBAL DEFAULT    9 _init
    18: 0000000000000a04     0 FUNC    GLOBAL DEFAULT   12 _fini
```

xoviat:
Great!

jcfr:
Thanks again @xoviat 

Changes integrated in https://github.com/scikit-build/scikit-build/commit/8b15eff9f8fb2f3d2fdd892ef0946efbb330c6eb

PR #300, opened 2018-02-22
escape detected entries in FindPythonExtensions.cmake

opadron:
Handles the case where the entries detected in the Python script may contain semicolons.  The script now correctly escapes these entries for proper consumption by CMake.

Thanks to @xoviat for originally pointing out the bug!


xoviat:
@opadron As seen in the test results, this doesn't address the problem.

Issue #289, opened 2018-02-18
 Cache build engine


ghost:
The build engine should be cached. In fact, scikit-build should not call `cmake` but rather `cmake --build .` if cmake has already been run.

jcfr:
Closing. Fixed by https://github.com/scikit-build/scikit-build/pull/318

PR #301, opened 2018-02-24
Add support for cached generators

xoviat:
Closes gh-289.

### DIFF
--- a/skbuild/cmaker.py
+++ b/skbuild/cmaker.py
@@ -40,6 +40,22 @@ def pop_arg(arg, args, default=None):
     return args, val
 
 
+def get_cached_generator():
+    """Reads and returns the cached generator from the BUILD_DIR; returns None
+    if not found.
+    """
+    try:
+        cmake_generator = 'CMAKE_GENERATOR:INTERNAL='
+        with open(os.path.join(CMAKE_BUILD_DIR, 'CMakeCache.txt')) as fp:
+            for line in fp:
+                if line.startswith(cmake_generator):
+                    return line[len(cmake_generator):].strip()
+    except:
+        pass
+    
+    return None
+
+
 def _remove_cwd_prefix(path):
     cwd = os.getcwd()
 
--- a/skbuild/setuptools_wrap.py
+++ b/skbuild/setuptools_wrap.py
@@ -440,9 +440,16 @@ def setup(*args, **kw):  # noqa: C901
     try:
         cmkr = cmaker.CMaker()
         if not skip_cmake:
-            env = cmkr.configure(cmake_args,
-                                 cmake_source_dir=cmake_source_dir,
-                                 cmake_install_dir=skbuild_kw['cmake_install_dir'])
+            # skip the configure step for a cached build
+            generator_name = cmaker.get_cached_generator()
+            if generator_name is not None:
+                env = cmkr.platform.get_generator(generator_name).env
+            else:
+                env = cmkr.configure(
+                    cmake_args,
+                    cmake_source_dir=cmake_source_dir,
+                    cmake_install_dir=skbuild_kw['cmake_install_dir']
+                )
             cmkr.make(make_args, env=env)
     except SKBuildGeneratorNotFoundError as ex:
         sys.exit(ex)
--- a/.gitignore
+++ b/.gitignore
@@ -49,6 +49,7 @@ docs/_build
 # IDE junk
 .idea/*
 *.swp
+.vscode/*
 
 # build output (testing)
 skbuild/_cmake_test_compile/*
--- a/skbuild/platform_specifics/abstract.py
+++ b/skbuild/platform_specifics/abstract.py
@@ -62,6 +62,16 @@ class CMakePlatform(object):
         """
         return "cmake"
 
+    def get_generator(self, generator_name):
+        """Loop over generators and return the first that matches the given
+        name.
+        """
+        for default_generator in self.default_generators:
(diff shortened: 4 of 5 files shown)


jcfr:
Very nice :+1:

skbuild/setuptools_wrap.py
```diff
@@ -440,9 +440,15 @@ def setup(*args, **kw):  # noqa: C901
...
         cmkr = cmaker.CMaker()
         if not skip_cmake:
-            env = cmkr.configure(cmake_args,
-                                 cmake_source_dir=cmake_source_dir,
-                                 cmake_install_dir=skbuild_kw['cmake_install_dir'])
+            # skip the configure step for a cached build
+            env = cmkr.get_cached_env()
+            if env is None:
+                env = cmkr.configure(
+                    cmake_args,
```
jcfr: One last thing would be to serialize the `cmake_args` into a file and compare it didn't change.

May a function  `get_cached_cmake_args` could be added, then configuration would be done only if the ``env is None` or `cmake_args != cached_cmake_args`

xoviat:
@jcfr Can we put this in?

skbuild/constants.py
```diff
@@ -14,3 +14,4 @@
 CMAKE_BUILD_DIR = os.path.join(SKBUILD_DIR, "cmake-build")
 CMAKE_INSTALL_DIR = os.path.join(SKBUILD_DIR, "cmake-install")
 SETUPTOOLS_INSTALL_DIR = os.path.join(SKBUILD_DIR, "setuptools")
+CMAKE_ARGUMENTS_FILE = os.path.join(CMAKE_BUILD_DIR, "CMakeArguments.txt")
```
jcfr: `CMakeArguments.json` ?

skbuild/setuptools_wrap.py
```diff
@@ -444,10 +466,17 @@ def setup(*args, **kw):  # noqa: C901
     try:
         cmkr = cmaker.CMaker()
         if not skip_cmake:
-            env = cmkr.configure(cmake_args,
-                                 cmake_source_dir=cmake_source_dir,
-                                 cmake_install_dir=skbuild_kw['cmake_install_dir'],
-                                 languages=cmake_languages)
+            # skip the configure step for a cached build
+            env = cmkr.get_cached_env()
+            if env is None or cmake_args != _load_cmake_args():
```
jcfr: I suggest to abort mentioning that a different version of cmake is being used, or attempt re-configure and let CMake report the message

xoviat: The current approach is to re-configure, so I'll do that.

skbuild/setuptools_wrap.py
```diff
@@ -24,6 +24,11 @@
 except ImportError:
     from io import StringIO
 
+try:
+    from shutil import which
+except ImportError:
+    from distutils.spawn import find_executable as which
```
jcfr: Instead, you could relocated the `which` function currently present in the test helpers.
See https://github.com/scikit-build/scikit-build/blob/d5fa09fb28caddad73f7a7f141969cdda28266b4/tests/__init__.py#L272-L292

skbuild/setuptools_wrap.py
```diff
@@ -459,6 +464,9 @@ def setup(*args, **kw):  # noqa: C901
     # weight and when CMake is given multiple times a argument, only the last
     # one is considered, let's prepend the one provided in the setup call.
     cmake_args = skbuild_kw['cmake_args'] + cmake_args
+    
+    # Used to confirm that the cmake executable is the same
+    cmake_cmd = [which('cmake')] + cmake_args
```
jcfr: I still think we should ask cmake what is its version.

After doing `pip  install -U cmake` or updating the version on the system, the path would be the same

xoviat:
@jcfr Wouldn't the version string be enough, even if CMake were on the path?

xoviat:
No, it wouldn't.

jcfr:
>Wouldn't the version string be enough, even if CMake located differently on the path?

I think the version string alone would be sufficient. I don't think we should check for the path.

Here code getting the version that could be refactored. 
https://github.com/scikit-build/scikit-build/blob/d5fa09fb28caddad73f7a7f141969cdda28266b4/skbuild/cmaker.py#L76-L80

xoviat:
I checked the `build.ninja`, and it has the full path to the CMake executable.

jcfr:
Good point. Also any [add_custom_command](https://cmake.org/cmake/help/latest/command/add_custom_command.html) using [CMAKE_COMMAND](https://cmake.org/cmake/help/latest/variable/CMAKE_COMMAND.html) would have the full path too, it means we need to use `which` as you started to do, and also check for the version

skbuild/cmaker.py
```diff
@@ -74,11 +74,15 @@ class CMaker(object):
...
         # verify that CMake is installed
         try:
-            subprocess.check_call(['cmake', '--version'])
+            version_string = subprocess.check_output(['cmake', '--version'])
         except (OSError, subprocess.CalledProcessError):
             raise SKBuildError(
                 "Problem with the CMake installation, aborting build.")
 
+        if sys.version_info > (3, 0):
+            version_string = version_string.decode()
```
xoviat: I have no idea what happens on PY2. @jcfr Ideas?

jcfr:
Closing. Superseded by https://github.com/scikit-build/scikit-build/pull/318

PR #333, opened 2018-07-09
travis: Remove linux testing, test on macos using Python 3.5, 3.6 and 3.7

jcfr:


jcfr:
Merging. After inspecting current state of the CI, issues associated with previous commit have been fixed.

PR #334, opened 2018-07-12
Use utf-8 to open cmake files

bgermann:
On Windows the default encoding for opening files is not utf-8.
This changes the scanning of .cmake files to open with utf-8.
If there is any character in a file that is not available in
CP-1252 the build fails.

### DIFF
--- a/skbuild/cmaker.py
+++ b/skbuild/cmaker.py
@@ -5,6 +5,7 @@ This module provides an interface for invoking CMake executable.
 import argparse
 import distutils.sysconfig as du_sysconfig
 import glob
+import io
 import itertools
 import os
 import os.path
@@ -414,7 +415,7 @@ class CMaker(object):
                 if os.path.splitext(filename)[1] != ".cmake":
                     continue
 
-                for line in open(os.path.join(root, filename), encoding="utf-8"):
+                for line in io.open(os.path.join(root, filename), encoding="utf-8"):
                     match = RE_FILE_INSTALL.match(line)
                     if match is None:
                         continue
--- a/skbuild/cmaker.py
+++ b/skbuild/cmaker.py
@@ -414,7 +414,7 @@ class CMaker(object):
                 if os.path.splitext(filename)[1] != ".cmake":
                     continue
 
-                for line in open(os.path.join(root, filename)):
+                for line in open(os.path.join(root, filename), encoding="utf-8"):
                     match = RE_FILE_INSTALL.match(line)
                     if match is None:
                         continue


jcfr:
Thanks @bgermann for the PR and for fixing the python 2 support :+1:

jcfr:
* [ ] To prevent future regression, I suggest you also add a test :smile: 
* [ ] Adding an entry to the [CHANGES.rst](https://github.com/scikit-build/scikit-build/blob/master/CHANGES.rst)

Inspiring from [this test](https://github.com/scikit-build/scikit-build/blob/ed72d417198af0d07cbd3a9148aab14998d75fc0/tests/test_cmaker.py#L69-L94), could you add a test named `test_issue334_configure_cmakelist_non_cp1252_encoding` (a similar name).

For example it could be similar to this (but you would still need to make sure the written `CMakeLists.txt`) is effectively failing without your patch:

```python

def test_issue334_configure_cmakelist_non_cp1252_encoding(configure_with_cmake_source_dir):
    tmp_dir = _tmpdir('test_issue334_configure_cmakelist_non_cp1252_encoding')
    with push_dir(str(tmp_dir)):
        tmp_dir.join('CMakeLists.txt').write(textwrap.dedent(
            """
            cmake_minimum_required(VERSION 3.5.0)
            project(foobar NONE)
            # Todo: Add here some problematic characters
            install(CODE "execute_process(COMMAND \${CMAKE_COMMAND} -E sleep 0)")
            """
        ))
        cmkr = CMaker()
        cmkr.configure()
```

If writing the `CMakeLists.txt` from within the test is too cumbersome, adding a sample project in [tests/samples](https://github.com/scikit-build/scikit-build/tree/master/tests/samples) is also possible (name for the project could be `issue334-configure-cmakelist-non-cp1252-encoding`). In that case, the test would look like this:

```python
def test_issue334_configure_cmakelist_non_cp1252_encoding(configure_with_cmake_source_dir):
    tmp_dir = _tmpdir('test_issue334_configure_cmakelist_non_cp1252_encoding')
    prepare_project('issue334-configure-cmakelist-non-cp1252-encoding', tmp_dir)
    with push_dir(str(tmp_dir)):
        cmkr = CMaker()
        cmkr.configure()
```

Ps: I haven't tried the suggest sample ... they may fail at the first try.

Ps2: To run the test, you can invoke `pytest test_cmaker.py::test_configure_cmakelist_non_cp1252_encoding::test_issue334_configure_cmakelist_non_cp1252_encoding`

Let me know if you have any questions, 
Jc

jcfr:
Closing. Superseded by #338

Issue #342, opened 2018-07-19
[OSX] skbuild.setup defaults cmake_args without checking kwargs for the arguments.


yonip:
OS: OSX/Darwin (testing on travis).

`setup` checks the command line arguments for `-DCMAKE_OSX_DEPLOYMENT_TARGET` and for `-DCMAKE_OSX_ARCHITECTURES`, and if they don't exist, it defaults them [here](https://github.com/scikit-build/scikit-build/blob/d8e4db57e0a930db75bd204e47b64ba83a7300bb/skbuild/setuptools_wrap.py#L461-L477). The issue with this is that if the user passed either of those arguments to setup via a kwarg. i.e. `setup(..., cmake_args=[..., "-DCMAKE_OSX_DEPLOYMENT_TARGET:STRING=10.7"])`, it gets overwritten with the default, even when nothing had been passed to the command line. 
This causes builds to break for no apparent reason (see the reasons for [this PR](https://github.com/skvark/opencv-python/pull/110)). A simple fix is moving the lines where the command line arguments are added to the arguments passed via kwarg from [here](https://github.com/scikit-build/scikit-build/blob/d8e4db57e0a930db75bd204e47b64ba83a7300bb/skbuild/setuptools_wrap.py#L503-L506) to right above the block mentioned above. 
Since `cmake_args` is not referenced between line 477 and line 503, nothing will get more args than it expected, and CLI will still override the args passed via kwarg, but default will only be set if neither the CLI or kwarg `cmake_arg` are set.
If this is approved I will [PR the fix](https://github.com/yonip/scikit-build/commit/2546c18c246857679e469b2a84576e1f802baa65) described above. 
(This is an issue first because the contribution guidelines specify filing bugs before trying to fix them).

jcfr:
Thanks for reporting the issue :+1: 

I will review and provide feedback later today.

yonip:
I have found another bug, in [ `Cmaker.has_cmake_cache_arg`](https://github.com/scikit-build/scikit-build/blob/master/skbuild/cmaker.py#L60-L70), which would cause the tests to fail: since it traverses the list of parameters in order, `has_cmake_cache_arg(['-DOVERRIDE:STRING=A', '-DOVERRIDE:STRING=B'], 'OVERRIDE', 'B')` returns `False`, while the way command line arguments are parsed, `OVERRIDE` should be set to `'B'`.
[...]

PR #344, opened 2018-07-25
Fix cmaker parse

yonip:
Fix cmaker parsing command line arguments in the wrong order when verifying them, as reported in followup to #342.


jcfr:
Will integrate after the CI completes

PR #346, opened 2018-07-26
Fix #342

### LINKED ISSUE #342: [OSX] skbuild.setup defaults cmake_args without checking kwargs for the arguments.

yonip:
Finally the PR with a test. 
Rebase did something weird with the commits, so if it matters I can probably reset and try again.

### DIFF
--- a/skbuild/setuptools_wrap.py
+++ b/skbuild/setuptools_wrap.py
@@ -458,6 +458,11 @@ def setup(*args, **kw):  # noqa: C901
         for parent_dir, file_list in kw.get('data_files', [])
     }
 
+    # Since CMake arguments provided through the command line have more
+    # weight and when CMake is given multiple times a argument, only the last
+    # one is considered, let's prepend the one provided in the setup call.
+    cmake_args = skbuild_kw['cmake_args'] + cmake_args
+
     if sys.platform == 'darwin':
 
         if plat_name is None:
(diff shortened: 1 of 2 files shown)


jcfr:
Thanks for the PR :+1: , no problem about all the commit. I will review and clean things up later this evening.

> re-linting

running `flake8` locally should help catch these before hand.

jcfr:
Explicitly killing AppVeyor since Travis failed ...there is no need to wait for windows (they are all greens)

jcfr:
@yonip I rebased and squashed

Let me know if you need help updating you local checkout

jcfr:
Closing. This PR is superseded by https://github.com/scikit-build/scikit-build/pull/349

PR #348, opened 2018-07-26
coverage: Fix "include-ignored" warning

jcfr:
This commit fixes the following warning:

  --include is ignored because --source is set (include-ignored)


jcfr:
Do not wait for appveyor.

PR #350, opened 2018-07-30
setuptools_wrap: Support package_dir with ending slash

jcfr:


jcfr:
Travis is now green - the last commit fixed it. And since most of the AppVeyor build are also green, I am merging.

PR #351, opened 2018-07-30
doc: Add advanced usage section "Enabling parallel build"

### LINKED ISSUE #61: Enable parallel build if support by underlying build tool
- If Makefile Unix, Ninja or Jom based generator are used:  `-j<number_of_core>`
- If Visual Studio:
  - `/MP<number_of_core>` 
  - `/maxcpucount:<number_of_core>` 

Note that for complex project, it could be hard to find the right setting for `/MP` and `/maxcpucount`. For example, you want to avoid too many link steps requiring a lot of memory to happen at the same time ....  Having a way to disable parallel build may be a sensible option.

More details:
- With VS >= 2010, target level parallelism can be set from command line using  `/maxcpucount`. (Number of simultaneous MSBuild.exe processes). See [1]
- object level parallelism set using `/MP`   (Number of simultaneous cl.exe processes). See [2]

[1] https://msdn.microsoft.com/en-us/library/bb651793%28v=vs.120%29.aspx
[2] https://msdn.microsoft.com/en-us/library/bb385193.aspx

Here is some more details of what we did for Slicer: http://massmail.spl.harvard.edu/public-archives/slicer-devel/2016/019656.html

jcfr:
Fixes #61

[ci skip]

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -277,6 +277,86 @@ can be set using the following approach::
         setup_requires.append('cmake')
 
 
+.. _usage_enabling_parallel_build:
+
+Enabling parallel build
+-----------------------
+
+Ninja
+^^^^^
+
+If :ref:`Ninja` generator is used, the associated build tool (called ``ninja``)
+will automatically parallelize the build based on the number of available CPUs.
+
+To limit the number of parallel jobs, the build tool option ``-j N`` can be passed
+to ``ninja``.
+
+For example, to  limit the number of parallel jobs to `3`, the following could be done::
+
+    python setup bdist_wheel -- -- -j3
+
+For complex projects where more granularity is required. It is also possible to limit
+the number of simultaneous link jobs, or compile jobs, or both.
+
+Indeed, starting with CMake 3.11, it is possible to configure the project with these
+options:
+
+* `CMAKE_JOB_POOL_COMPILE <https://cmake.org/cmake/help/latest/variable/CMAKE_JOB_POOL_COMPILE.html>`_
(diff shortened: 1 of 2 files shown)


docs/usage.rst
```diff
@@ -277,6 +277,86 @@ can be set using the following approach::
...
+
+If :ref:`Ninja` generator is used, the associated build tool (called ``ninja``)
+will automatically parallelize the build based on the number of available CPUs.
+
+To limit the number of parallel jobs, the build tool option ``-j N`` can be passed
+to ``ninja``.
+
+For example, to  limit the number of parallel jobs to `3`, the following could be done::
+
+    python setup bdist_wheel -- -- -j3
```
thewtex: `setup` -> `setup.py`

docs/usage.rst
```diff
@@ -277,6 +277,86 @@ can be set using the following approach::
...
+
+For complex projects where more granularity is required. It is also possible to limit
```
thewtex: *required. It* -> *required, it*

docs/usage.rst
```diff
@@ -277,6 +277,86 @@ can be set using the following approach::
...
+Unix Makefiles
+^^^^^^^^^^^^^^
+
+If :ref:`Unix Makefiles` generator is used, the associated build tool (called ``make``)
+will **NOT** automatically parallelize the build, the user has to explicitly pass
+option like ``-j N``.
+
+For example, to limit the number of parallel jobs to `3`, the following could be done::
+
+    python setup bdist_wheel -- -- -j3
```
thewtex: setup -> setup.py

jcfr:
Thanks for the review

PR #366, opened 2018-11-15
Fix #365 for the case of compiling with Hombrew GCC

### LINKED ISSUE #365: Linker Error when building with brewed g++ on macOS
**Operating System**: macOS Mojave 10.14.1
**Python Version**: 3.7
**C Compiler**:  Homebrew GCC 8.2.0 

**Issue**:
Related to issue #362, I am building a project using Homebrew GCC. However, the Cython build step fails at the linking stage with `ld: unknown option: --version-script=...` 

**Possible Resoltion* The version-script command is passed from `skbuild/resources/cmake/FindPythonExtensions.cmake` when `"${CMAKE_C_COMPILER_ID}" STREQUAL "GNU"`, so we can add an additional condition that the platform is not Darwin. This fixes compilation in my case.

padraic-padraic:
As described in #365 this is a (slightly hacky) fix for compilation with Homebrew GCC and Python 3.

Not sure why the tests failed; flake8 is reporting errors with files I didn't touch.

### DIFF
--- a/skbuild/resources/cmake/FindPythonExtensions.cmake
+++ b/skbuild/resources/cmake/FindPythonExtensions.cmake
@@ -332,7 +332,7 @@ function(_set_python_extension_symbol_visibility _target)
     set_target_properties(${_target} PROPERTIES LINK_FLAGS 
         "/EXPORT:${_modinit_prefix}${_target}"
     )
-  elseif("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU")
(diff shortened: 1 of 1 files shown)


jcfr:

> Not sure why the tests failed; flake8 is reporting errors with files I didn't touch.

This should be fixed in https://github.com/scikit-build/scikit-build/pull/368

> fix for compilation with Homebrew GCC and Python 3

I will review and integrate shortly.

jcfr:
Topic rebased to include style fix.

thewtex:
Azure Pipelines spins up a pretty consistent VM, but I am not sure if it would work better than AppVeyor in this case.

thewtex:
> This failed because of 1hr timeout.

That said, Azure Pipelines has a 6 hour timeout :-).

PR #377, opened 2019-01-04
Allow for setting OSX plat_name from cmake args

nonhermitian:
It is possible to change the target OSX platform using `--plat-name macosx-10.x-x86_64` or via the cmake args `'-DCMAKE_OSX_DEPLOYMENT_TARGET:STRING=10.x', '-DCMAKE_OSX_ARCHITECTURES:STRING=x86_64'`.  However, only the former sets the platform correctly for output wheels, e.g. via `bdist_wheel`.  Therefore its is possible to build for `macosx-10.9-x86_64`, as required for the latest version of OSX, but have the output wheel still say `macosx-10.6-x86_64`.  

This PR sets the `plat_name` for OSX from the cmake args if they are explicitly set by the user, and propagates this to the later build commands by appending to the `sys.argv` so that it behaves in the same manner as doing `--plat-name macosx-10.x-x86_64`

### DIFF
--- a/skbuild/setuptools_wrap.py
+++ b/skbuild/setuptools_wrap.py
@@ -474,8 +474,31 @@ def setup(*args, **kw):  # noqa: C901
     # specified
     if sys.platform == 'darwin':
         if plat_name is None:
-            # The following code is duplicated in bdist_wheel.finalize_options()
-            plat_name = "macosx-10.6-x86_64"
+            # If cmake osx deployment args are set then set plat_name from them
+            user_set = False
+            if cmaker.has_cmake_cache_arg(
+                    cmake_args, 'CMAKE_OSX_DEPLOYMENT_TARGET'):
+                version = "".join(s for s in cmake_args
+                                  if 'CMAKE_OSX_DEPLOYMENT_TARGET'
+                                  in s).split('=')[1]
+                user_set = True
+            else:
+                version = '10.6'
+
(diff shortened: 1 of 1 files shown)


skbuild/setuptools_wrap.py
```diff
@@ -474,8 +474,33 @@ def setup(*args, **kw):  # noqa: C901
...
+            if cmaker.has_cmake_cache_arg(
+                    cmake_args, 'CMAKE_OSX_DEPLOYMENT_TARGET'):
+                # The loop here allows for cli flags to overload
+                # those set in the setup file.
+                for s in cmake_args:
+                    if 'CMAKE_OSX_DEPLOYMENT_TARGET' in s:
+                        version = s.split('=')[1]
+                user_set = True
+            else:
+                version = '10.6'
```
thewtex: Now-a-days, I think `10.9` is more common, per

https://github.com/MacPython/wiki/wiki/Spinning-wheels

thewtex:
@nonhermitian this is great!

@jcfr there seems to be some issue with the Python 3.5 / Windows build?

atilag:
Looking forward to seeing this PR merged!

jcfr:
>  there seems to be some issue with the Python 3.5 / Windows build?

The failure is due to appveyor timeout ... :(

jcfr:

jcfr:
Closing. This is superseded by #380

PR #378, opened 2019-01-05
Add helper API

Erotemic:
As a scikit, scikit-build should strive to make its target area of science simpler. However, unlike other scikits like learn and image, there isn't much of a top-level API. Designing a good top-level "build" API for Python is something that doesn't yet exist. Scikit-build has an opportunity to be the first tool that actually makes writing a `setup.py` script simple.

As such I wanted to propose a few baseline functions to populate the currently singular API. These functions are contained in a file `helper_api.py`. The top of that file contains a docstring detailing a use-case that illustrates the usefulness of these API calls. All of these API calls serve the purpose of making it easier to fill out arguments to `setuptools.setup`.

The API calls are:

* find_packages - a port from setuptools to populate `packages`
* parse_version - parse versions with static analysis and populate `version`
* parse_long_description - for populating `long_description` easier
* parse_requirements - for populating `install_requires` easier
* parse_authors - basic parsing of authors from git

I've currently been copy+pasting these 200ish lines of code between all of my repos, but it seems to me that scikit-build might be the right place for this logic to actually live. 

I've exposed the functions in the `utils` subpackage. I'm not sure where these are best placed in the directory structure, so I'm open to ideas / suggestions here.


Erotemic:
I've updated these functions to be slightly simpler and I moved them to `skbuild.utils`, which I think makes more sense from a namespace perspective. 

I'm also trying out a slightly expanded version of these tools in another PR: 
https://github.com/mariusmuja/flann/blob/a8e9f53040f7f01d3e915e4c9669c47428bee1fb/skbuild_template.py This includes additional helpers for working with classifiers, licenses, and shared library extensions. I'm not sure if they are general enough to warrant inclusion, so I've opted to not include them here by default. However, I'm open to maintainer input on this.

henryiii:

Issue #352, opened 2018-08-01
Fix support for isolated build


jcfr:
The problem we are trying to solve is the following:

With the latest pip, trying to run:

```
pip install -e .
```

twice in a row led to an error like the following:

```
    -- Python install dir is /home/jcfr/.virtualenvs/scikit-build-35/lib/python3.5/site-packages
    CMake Error at src/python/CMakeLists.txt:68 (find_package):
      By not providing "FindPythonExtensions.cmake" in CMAKE_MODULE_PATH this
      project has asked CMake to find a package configuration file provided by
      "PythonExtensions", but CMake did not find one.

      Could not find a package configuration file provided by "PythonExtensions"
      with any of the following names:

        PythonExtensionsConfig.cmake
        pythonextensions-config.cmake

      Add the installation prefix of "PythonExtensions" to CMAKE_PREFIX_PATH or
      set "PythonExtensions_DIR" to a directory containing one of the above
      files.  If "PythonExtensions" provides a separate development package or
      SDK, be sure it has been installed.
```

Indeed, during first configuration the `CMAKE_MODULE_PATH` cache variable was set to to `/tmp/pip-build-env-2_kd457b/lib/python3.5/site-packages/skbuild/resources/cmake` which point to the CMake module from the isolated build environment.

jcfr:
Waiting this is addressed, this can be mitigated running:

```
pip install -e . --no-build-isolation
```

jcfr:
ref #124

PR #379, opened 2019-01-06
Add support for build_ext and --inplace

Erotemic:
This PR addresses issues (#284 and #352) where command `python setup.py build_ext --inplace` fails when using scikit-build. To allow for running `build_ext`, I wrote a command class following the template I saw in `build` and `build_py`. Unfortunately, this didn't just make the normal `--inplace` option work because the `setuptools.**.build_ext` command only copies registered extension modules, which the cmake outputs are not. To remedy this I simply hacked in a check for `'--inplace' in sys.argv` and then copied the libraries to the source directory as would be done in "developer_mode". 

I'm not 100% this is the best way to implement the feature as (1) setuptools is complex and (2) I'm still getting familiar with how this library works and its standards.

### DIFF
--- a/skbuild/command/build_ext.py
+++ b/skbuild/command/build_ext.py
@@ -1,13 +1,47 @@
 """This module defines custom implementation of ``build_ext`` setuptools command."""
 
+import os
+
+from distutils.file_util import copy_file
+
 try:
     from setuptools.command.build_ext import build_ext as _build_ext
 except ImportError:
     from distutils.command.build_ext import build_ext as _build_ext
 
 from . import set_build_base_mixin
+from ..constants import CMAKE_INSTALL_DIR
 from ..utils import new_style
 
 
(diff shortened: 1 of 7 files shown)


jcfr:
Thanks for the contributions :+1: 

Could you clarify how the PR is related to issue https://github.com/scikit-build/scikit-build/issues/352 ? 

I will review and integrate shortly

Erotemic:
Thanks for the review!

Perhaps it's not related to #352, I was quickly scanning issues and PRs to make sure I wasn't duplicating work, and I found #284, which is related. That issue references #352, and at a quick glance I must have thought it seemed relevant.

jcfr:
`tests/test_issue284_build_ext_inplace.py::test_build_ext_inplace_command` is passing for both Python 3.5 and 3.7

PR #380, opened 2019-01-09
Ensure wheel platform name is correct when providing custom CMAKE_OSX_* values

jcfr:
Supersede #377


jcfr:
@nonhermitian Thanks again for your contribution. 

@nonhermitian  @atilag Would appreciate if you could review :+1:

jcfr:
This topic is now ready for final review with all issues addressed. Once CI returns green, it will be integrated.

PR #385, opened 2019-01-30
Sync setup with dev requirments

FlorianFranzen:
Currently ```setup.py``` contains an outdated requirements expression for ```pytest-runner```, which breaks testing with newer versions of the dependency.

This pull request updates this requirement based on the current ```requirements-dev.txt```.


jcfr:
@FlorianFranzen  Thanks again. I also updated the CHANGES.rst to reference your contribution :smile: . See https://github.com/scikit-build/scikit-build/commit/9566de5fda583c830312be0ffab2dcb0256ecd88

PR #393, opened 2019-03-27
fix test with system wide install of "Visual C++ for Python 2.7"

jcfr:


jcfr:
Merging. Build on scikit-build-feedstock confirmed it worked

image

See https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=25015

PR #400, opened 2019-04-19
Test generator candidates with call(list), not str

jokva:
Instead of building a command string for subprocess.call, build the
arguments as a list. That way, subprocess properly handles spaces in
arguments and paths.

The problem with the string-based invocation of subprocess is uses of
skbuild such as

  setup.py build --cmake-executable "C:/Program Files/cmake.exe" ..

fails awkwardly (without actually stopping configuration) with the error
output:

'C:/Program' is not recognized as an internal or external command,
operable program or batch file.

This does not occur when the executable and arguments are passed as a
list.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,11 @@ New Features
 Bug fixes
 ---------
 
+* Checking generator candidates can now handle handle paths and binaries with
+  spaces, so that ``setup.py --cmake-executable "C:/Program Files
+  (x86)/cmake/cmake.exe"`` works as expected.
+  Contributed by :user:`jokva`. See :issue:`400`.
+
 * Fix sdist command to ensure symlinks in original source tree are maintained.
   Contributed by :user:`anibali`. See :issue:`401`.
 
--- a/skbuild/platform_specifics/abstract.py
+++ b/skbuild/platform_specifics/abstract.py
@@ -170,9 +170,6 @@ class CMakePlatform(object):
         # Do not complain about unused CMake arguments
         cmake_args.insert(0, "--no-warn-unused-cli")
 
-        cmake_args_as_str = " ".join(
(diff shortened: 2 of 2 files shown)


jokva:
A consequence here is that shell = True is removed. I don't know if there should be any problems with that (it isn't used for any of the other cmake invocations), but I leave that up to you maintainers.

jokva:
Looks like the tests fail _because_ shell = False, my guess because they rely on environment variable substitution for stuff with env (maybe CMAKE_MAKE_PROGRAM). It's not clear from the subprocess docs, but I guess it relies on the shell to actually update the environment.

jokva:
Looks like the env thing might've been a faulty CI instance. Some windows install tests still fails, but I can't really see why they'd come from this patch. Any ideas?

jcfr:
Thanks for the PR :pray:  :+1: 

Now, the CI seems to only be failing only because the CODECOV_TOKEN is not available to PR from forks. Nothing critical.

Before we integrate, could you add an entry to https://github.com/scikit-build/scikit-build/blob/master/CHANGES.rst

jokva:
Of course, coming right up!

It's slightly difficult to test this as it relies on there (always?) being spaces in the argument, in particular to the executable itself. Even on Windows there's no guarantee for that happening as the used cmake can be in C:/bin/cmake/cmake.exe. Should we pursue a test for this specific case, or just be happy with the test suite otherwise passing?

Maybe on appveyor the path does include spaces, but the test would be rather fragile anyway.

jokva:
I don't know if it was clear because of the force push, but I added the changelog entry, so as far as i can tell this is ready and done. If there is more to do, please let me know.

jcfr:
Thanks again for your patience. Changes integrated in https://github.com/scikit-build/scikit-build/commit/6c0ed7b08e25c6a7267b44f43a436d818009b41b

PR #404, opened 2019-05-10
test: Fix reporting of assert associated with check_wheel_content

jcfr:
To allow pytest to display useful information, the tests containing
the helper function is registered.

See https://docs.pytest.org/en/latest/writing_plugins.html#assertion-rewriting


jcfr:
We may have to start pinning dependencies to avoid issue like this one:
* 1 hours ago ... tests checking wheel content were passing, but the one checking sdist content were failing
* In the mean, I fixed the test checking content of the sdist  :v: 
* but the wheel one start to fail :astonished: 

Humm ... :thinking: 

And here is the culprit ... 

image


image

PR #407, opened 2019-05-13
ENH: flexibly encode the implementation

mattip:
The interpreter implementation is hard coded, this PR at least extends the range of hard-coded values. It would be even better to read this out somewhere. `platform.python_implementation()` would get a unique name, maybe the possible values could be mapped to a canonical implementation name.

Right now PyPy's is `pypy-c`, but I put in `pypy` as well to maybe future-proof the logic.

With this change, and a few small tweaks to PyPy's sysconfig `'LIB` and `LIBDIR` values, I can successfully build opencv-python (which uses scikit-build) on pypy2.7

### DIFF
--- a/skbuild/cmaker.py
+++ b/skbuild/cmaker.py
@@ -315,9 +315,10 @@ class CMaker(object):
                 candidate_versions += ('',)
 
             candidates = (
-                os.path.join(prefix, ''.join(('python', ver)))
-                for (prefix, ver) in itertools.product(
+                os.path.join(prefix, ''.join((impl, ver)))
+                for (prefix, impl, ver) in itertools.product(
                     candidate_prefixes,
+                    ('python', 'pypy', 'pypy-c'),
                     candidate_versions
                 )
             )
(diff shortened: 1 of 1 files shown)


jcfr:

This is great, all test pass  (the failure is unrelated and due to independent issue that will be addressed shortly)

skbuild/cmaker.py
```diff
@@ -381,10 +382,11 @@ def get_python_library(python_version):
             candidates = (
                 os.path.join(
                     libdir,
-                    ''.join((pre, 'python', ver, abi, ext))
+                    ''.join((pre, impl, ver, abi, ext))
                 )
-                for (pre, ext, ver, abi) in itertools.product(
+                for (pre, impl, ext, ver, abi) in itertools.product(
                     candidate_lib_prefixes,
+                    ('python', 'pypy', 'pypy-c'),
```
jcfr: What do you think of using a global dictionary and lookup the implementation based on the return value of `platform.python_implementation()` ? 

It would also be nice to setup CI for testing scikit-build with PyPy ?  Is it something you could help  with ? May be extending the existing CircleCI or azure-pipelines configuration ?

mattip: Which one would you prefer, circleci or azure?

jcfr: If you end up using a Docker image, leveraging CircleCI may make more sense as you can easily specify a container to run the test with.

May be you could look into using one of these images: https://marketplace.automic.com/details/pypy-official-docker-image ?

On linux, I wonder if requirements for pypy are similar to the one for manylinux ? (meaning a compiling environment with a modern compiler but dependency on an older  glibc. See [pep-513](https://www.python.org/dev/peps/pep-0513) and [pep-571](https://www.python.org/dev/peps/pep-0571))

After the details are ironed out, I also wonder if (similarly to what is done for manylinux) integrating an image in [dockcross](https://github.com/dockcross/dockcross) would be relevant. This would provide one (or multiple images) with a readily available compiling env.

jcfr:
Closing. This is superseded by https://github.com/scikit-build/scikit-build/pull/418

PR #402, opened 2019-05-09
Use sdist command from setuptools instead of distutils

anibali:
Fixes https://github.com/scikit-build/scikit-build/issues/401.

Notes:
* Updated `generate_source_manifest.py` to produce a `MANIFEST.in` file instead of a `MANIFEST` file.

Issue #401, opened 2019-05-09
Undesirable behaviour with sdist and symlinks


anibali:
When using `sdist` with setuptools, symlinks are replaced with the concrete files that they reference. I find this useful for keeping a version number consistent across multiple projects in the same repo.

In contrast, when using `sdist` with skbuild, the symlink is not followed, resulting in a (usually) useless symlink being added to the archive. This has proven to be a rather large annoyance in my current attempt at migrating a project over to scikit-build builds.

I have created a minimal example at https://github.com/anibali/issue-skbuild-sdist to exhibit both behaviours.

Steps to reproduce:

1. Clone https://github.com/anibali/issue-skbuild-sdist.
2. Run `SETUPTOOLS=1 python setup.py sdist` to use setuptools' `setup`.
3. Look at the contents of the produced tarball. The VERSION file will be a concrete file.
4. Run `python setup.py sdist` to use skbuild's `setup`.
5. Look at the contents of the produced tarball. The VERSION file will be a useless symlink.

Would using setuptools' sdist in scikit-build's sdist command be a possibility?: https://github.com/scikit-build/scikit-build/blob/master/skbuild/command/sdist.py#L6

jcfr:
Thanks for the details report and the associated test case.

I was able to reproduce the problem.

> Would using setuptools' sdist in scikit-build's sdist command be a possibility?

Definitively, setuptools being a requirement, that makes sense.

I confirmed that this was working. Could you create a PR ? It may be sensible to  also add a small regression  test similar to  the existing [ones](https://github.com/scikit-build/scikit-build/tree/master/tests) (we have prefix `test_issueXXX`)

```diff
diff --git a/skbuild/command/sdist.py b/skbuild/command/sdist.py
index aa3230a..a55eea7 100644
--- a/skbuild/command/sdist.py
+++ b/skbuild/command/sdist.py
@@ -3,7 +3,7 @@
 import os
 
 from distutils import log as distutils_log
-from distutils.command.sdist import sdist as _sdist
+from setuptools.command.sdist import sdist as _sdist
 
 from . import set_build_base_mixin
 from ..utils import distribution_hide_listing, new_style

```

anibali:
Sure, I'll work on putting together a PR soon.

anibali:
[...]

PR #408, opened 2019-05-13
Improve sdist command to ensure symlinks are maintained.

jcfr:
This commit backports _remove_os_link() context manager
from setuptools.sdist and re-implement make_distribution()
to ensure symlinks in source tree are maintained in generated
source distribution (sdist).

It addresses issue #401.

Note that simply updating the sdist implementation of scikit-build
to extend from the setuptools one instead of the distutils one
would have other side-effect requiring to revisit the implementation
of generate_source_manifest() custom command. For more details, see #402

Co-authored-by: Jean-Christophe Fillion-Robin <jchris.fillionr@kitware.com>


jcfr:
azure-pipelines failure is related to codecov submission and can be ignored for now

PR #281, opened 2018-02-16
Add UseF2PY

xoviat:

PR #413, opened 2019-05-14
Add UseF2PY

jcfr:
Update: this function in only handling pyf files, otherwise, it becomes too complex.

This PR supersedes https://github.com/scikit-build/scikit-build/pull/281 originally submitted by `@xoviat`


bnavigator:
Please look at #449 before merging this.

PR #414, opened 2019-05-14
UseCython: attempt to handle generated dependencies

### LINKED ISSUE #292: add_cython_target fails to account for generated Cython files
Suppose you generate a file with `add_custom_command` and then pass the generated file into `add_cython_target`. The current function fundamentally fails to account for this situation. I've attempted multiple times to resolve it, but have not been successful.

The following issues arise:
1. The function cannot detect pxd dependencies (for obvious reasons), so I added a new DEPENDS argument where you can specify them.
2. Because the file is added in the build directory, the include directories break horribly. I have not been able to get around this issue.

jcfr:
- Add ability to read C or CXX from source properties
- Add DEPENDS to Cython command
- Use --fast-fail to improve build output
- Improve handling of generated Cython files

Note: the include directory fix is ugly, but Cython requires that `__init__.py` files are present in the build directory to read the `.pxd` files, and I haven't identified a better fix. I really would prefer that someone else fixed this problem because this solution feels icky to write.

Closes gh-292.

This PR supersedes https://github.com/scikit-build/scikit-build/pull/298 originally submitted by `@xoviat`

### DIFF
--- a/skbuild/resources/cmake/UseCython.cmake
+++ b/skbuild/resources/cmake/UseCython.cmake
@@ -205,15 +205,22 @@ function(add_cython_target _name)
 
   set(comment "Generating ${_output_syntax} source ${generated_file_relative}")
   set(cython_include_directories "")
-  set(pxd_dependencies "")
+  set(pxd_dependencies "${args_DEPENDS}")
   set(c_header_dependencies "")
+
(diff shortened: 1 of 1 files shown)


henryiii:

PR #418, opened 2019-05-23
Add support for PyPy and setup continuous integration

mattip:
Let's see if this works.

Since PyPy is still in a catch-up mode where it releases every 3-6 months, I preferred to `wget` the latest release.

Follow on from #407, will fail until that is merged

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,12 +14,19 @@ New Features
 * Add support for passing :ref:`CMake configure options <usage_cmake_configure_options>` like `-DFOO:STRING:bar`
   as global `setuptools` or `pip` options.
 
-
 Bug fixes
 ---------
 
 * Fix unclosed file resource in :meth:`skbuild.cmaker.CMaker.check_for_bad_installs`.
   Thanks :user:`Nic30` for the suggestion. See :issue:`429`.
+
+Tests
+-----
+
+* Continuous Integration
+
+  * Update Azure Pipelines configuration for running tests using `PyPy3 <https://pypy.org/>`_ on Linux.
+    Thanks :user:`mattip` for the contribution. See :issue:`418`.
 
 Scikit-build 0.10.0
 ===================
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,9 @@ New Features
 * Add support for passing :ref:`CMake configure options <usage_cmake_configure_options>` like `-DFOO:STRING:bar`
   as global `setuptools` or `pip` options.
 
+* Add support for building project using PyPy or PyPy3. See https://pypy.org
+  Thanks :user:`mattip` for the contribution. See :issue:`407`.
+
 Bug fixes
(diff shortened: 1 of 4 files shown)


mattip:
It seems there is a problem installing cmake in the pre-install step, since it depends on skbuild (recursive dependency?):
```
[scikit-ci] Executing: pip install cmake
Collecting cmake
  Downloading https://files.pythonhosted.org/packages/72/7b/8d3d83f884283ddfc86ca407d10152d66b84dff5baf5ed4537c127f988cb/cmake-3.13.3.tar.gz
    ERROR: Complete output from command python setup.py egg_info:
    ERROR: Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-zcccb0ai/cmake/setup.py", line 7, in <module>
        from skbuild import setup
    ModuleNotFoundError: No module named 'skbuild'
    ----------------------------------------
ERROR: Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-install-zcccb0ai/cmake/
```

jcfr:
>  problem installing cmake

Indeed, cmake wheel are currently specific to the python abi. I think this will be addressed implementing the approach documented in https://github.com/scikit-build/scikit-build/issues/420 . 

Do you think that wheel with tag like `py2.py3-none-manylinux2010_x86-64` will be installable on pypy ?

mattip:
> Do you think that wheel with tag like py2.py3-none-manylinux2010_x86-64 will be installable on pypy

Yes, once I updated pip to version 19, this command `pypy3.6-v7.1.1/bin/pypy3 -mpip install imageio-ffmpeg` downloaded and installed `imageio_ffmpeg-0.3.0-py3-none-manylinux2010_x86_64.whl`

mattip:
Rebased off master to get the new cmake package.

The build failed. While watching it, it seemed to cycle infinitely through some pep-8 flake thing, checking all the site packages. Eventually I think it timed out.

When I try to reproduce the steps localy, I cannot run`python -mci install`. When I "lie" to it, by running `APPVEYOR=True python -m ci install`, the process escapes the virtual environment and `pip install cmake` tries to install it to my OS python2.

jcfr:
> it seemed to cycle infinitely through some pep-8 flake thing

The issue was that folders for pypy and associated environment were created in the scikit-build source directory and flake8 was trying to lint all the associated files. This is now fixed by creating directories in the parent folder.

>  When I "lie" to it, by running APPVEYOR=True python -m ci install

To "simulate" Azure Pipelines locally, the variables `TF_BUILD=true` and 'AGENT_OS=Linux' could be set.

jcfr:
It turns out that `get_python_include_dir` does not need to be modified. Indeed the header `Python.h` is also available within PyPy installation.

Regarding `get_python_library`, I fixed and simplified the list adding `pypy3-c`.

mattip:
cool, tests pass

Issue #383, opened 2019-01-10
Fix compiled extension for Python >= 3.5


jcfr:
Retrieving compiled extension suffix using

```
import sysconfig
sysconfig.get_config_var('SO')
```

is incorrect. Indeed, it always return `.pyd` but starting with Python 3.5, extension is expected to have the form `.cpXY-win_amd64.pyd` or `.cpXY-win32.pyd`

This was introduced in https://github.com/python/cpython/commit/65e4cb10d9d9964f30bc72561bf0e86833328a3b 

From `Lib/distutils/sysconfig.py`:
```
g['EXT_SUFFIX'] = _imp.extension_suffixes()[0]
```

jcfr:
Closing. Fixed in https://github.com/scikit-build/scikit-build/pull/433

Issue #381, opened 2019-01-10
DeprecationWarning: SO is deprecated, use EXT_SUFFIX


jcfr:
Use of `sysconfig.get_config_var('SO')` should be changed to `sysconfig.get_config_var('EXT_SUFFIX')` if supported by the python being used

https://github.com/scikit-build/scikit-build/blob/3c6bcf01704c74652198d4ced748d4801e71120b/skbuild/resources/cmake/FindPythonExtensions.cmake

https://github.com/scikit-build/scikit-build/blob/master/tests/test_hello_cpp.py

https://github.com/scikit-build/scikit-build/blob/master/tests/test_hello_cython.py

jcfr:
See #383

jcfr:
Closing. Fixed in https://github.com/scikit-build/scikit-build/pull/433

PR #433, opened 2019-06-12
FindPythonExtensions: Fix deprecated warning and use correct python extension suffix on windows

jcfr:
See #381 and #383


jcfr:
This should allow to remove this workaround:

https://github.com/fastscape-lem/fastscapelib-fortran/blob/a8b90ae78fd7e83af86dcf84e5853223477c3605/CMakeLists.txt#L140-L145

Cc: @benbovy

PR #441, opened 2019-07-26
Use the MACOSX_DEPLOYMENT_TARGET env variable

certik:
If the variable is defined, use it. Othewise use the value from distutils.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -29,6 +29,9 @@ Bug fixes
 
   * Fix warning using `EXT_SUFFIX` config variable instead of deprecated `SO` variable. See :issue:`381`.
 
+* Honor the `MACOSX_DEPLOYMENT_TARGET` environment variable if it is defined on
+  macOS. Thanks :user:`certik` for the contribution. See :issue`441`.
+
 Documentation
 -------------
 
--- a/skbuild/constants.py
+++ b/skbuild/constants.py
@@ -20,10 +20,16 @@ def _default_skbuild_plat_name():
     On macOS, it corresponds to the version and machine associated with :func:`platform.mac_ver()`.
     """
     if sys.platform == 'darwin':
-        # distutils.util.get_platform() returns the release based on the value
-        # of MACOSX_DEPLOYMENT_TARGET on which Python was built, which may
-        # be significantly older than the user's current machine.
+        # If the MACOSX_DEPLOYMENT_TARGET environment variable is defined, use
+        # it, as it will be the most accurate. Otherwise use the value returned
+        # by platform.mac_ver() provided by the platform module available in
+        # the Python standard library.
+        #
+        # Note that on macOS, distutils.util.get_platform() is not used because
+        # it returns the macOS version on which Python was built which may be
+        # significantly older than the user's current machine.
         release, _, machine = platform.mac_ver()
+        release = os.environ.get("MACOSX_DEPLOYMENT_TARGET", release)
         split_ver = release.split('.')
         return 'macosx-{}.{}-{}'.format(split_ver[0], split_ver[1], machine)
     else:
--- a/docs/generators.rst
+++ b/docs/generators.rst
@@ -266,6 +266,11 @@ If needed, this can be overridden by explicitly passing the CMake option
 Customizing Deployment Target and Architecture
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+.. versionadded:: 0.11.0
+
+Deployment target can be customized by setting the `MACOSX_DEPLOYMENT_TARGET`
+environment variable.
+
 .. versionadded:: 0.7.0
 
 Deployment target and architecture can be customized by associating the
--- a/tests/test_issue342_cmake_osx_args_in_setup.py
+++ b/tests/test_issue342_cmake_osx_args_in_setup.py
@@ -1,15 +1,21 @@
 import pytest
 import sys
 import textwrap
+import platform
 
 import skbuild.constants
 
-from . import _tmpdir, execute_setup_py
+from . import _tmpdir, execute_setup_py, push_env
+
+params = "osx_deployment_target_env_var,cli_setup_args," \
+    "keyword_cmake_args,cli_cmake_args,expected_cmake_osx_deployment_target"
 
 
-@pytest.mark.parametrize("cli_setup_args,keyword_cmake_args,cli_cmake_args,expected_cmake_osx_deployment_target", [
+@pytest.mark.parametrize(params, [
     # default plat_name is 'macosx-10.6-x86_64'
     (
+            # osx_deployment_target_env_var
+            None,
             # cli_setup_args
             [],
             # keyword_cmake_args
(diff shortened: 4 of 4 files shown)


skbuild/constants.py
```diff
@@ -20,10 +20,15 @@ def _default_skbuild_plat_name():
...
-        # be significantly older than the user's current machine.
+        # If the MACOSX_DEPLOYMENT_TARGET environment variable is defined, use
+        # it, as it will be the most accurate. Otherwise use the value returned
+        # by distutils. distutils.util.get_platform() returns the release based
+        # on the value of MACOSX_DEPLOYMENT_TARGET on which Python was built,
+        # which may be significantly older than the user's current machine.
         release, _, machine = platform.mac_ver()
+        mdt = os.getenv("MACOSX_DEPLOYMENT_TARGET")
+        if mdt:
+            release = mdt
```
jcfr: You could even do this ... 

```python
release = os.environ.get("MACOSX_DEPLOYMENT_TARGET", release)
```

skbuild/constants.py
```diff
@@ -20,10 +20,15 @@ def _default_skbuild_plat_name():
     On macOS, it corresponds to the version and machine associated with :func:`platform.mac_ver()`.
     """
     if sys.platform == 'darwin':
-        # distutils.util.get_platform() returns the release based on the value
-        # of MACOSX_DEPLOYMENT_TARGET on which Python was built, which may
-        # be significantly older than the user's current machine.
+        # If the MACOSX_DEPLOYMENT_TARGET environment variable is defined, use
```
jcfr: For the comment, I instead suggest something like this:
```
        # If the MACOSX_DEPLOYMENT_TARGET environment variable is defined, use
        # it, as it will be the most accurate. Otherwise use the value returned
        # by platform module from the Python standard library.
        # 
        # Note that since distutils.util.get_platform() returns the release based
        # on the value of MACOSX_DEPLOYMENT_TARGET on which Python was built,
        # (which may be significantly older than the user's current machine), it can not be used.
```

skbuild/constants.py
certik: Even better. Fixed in 698597d4712350887faef7b1c35d0425f8d88019.

skbuild/constants.py
certik: The very last sentence is confusing. Instead of

> (which may be significantly older than the user's current machine), it can not be used.

I would suggest

> (which may be significantly older than the user's current machine), it should only be used if the MACOSX_DEPLOYMENT_TARGET environment variable is not defined.

certik:
I updated the CHANGES. Would you mind helping me with the Documentation? Where do you think this should be mentioned?

Also, how would you suggest to test this? It's not quite clear to me from the test you mentioned.

skbuild/constants.py
jcfr: In fact there are two ways of getting the platforms:

```
import platform
print("approach 1", platform.mac_ver())

from distutils.util import get_platform
print("approach 2", get_platform())
```

On macOS, the approach1 is used and the comment `Note that since distutils.util.get_platform() ...` is here to explain why approach 2 is not used.

certik: That assumes that `MACOSX_DEPLOYMENT_TARGET` is always defined on macOS. I don't know if that's always necessarily the case. So the code should still work on macOS even if MACOSX_DEPLOYMENT_TARGET is not defined. 

I didn't know about [platform.mac_ver()](https://docs.python.org/3.8/library/platform.html#mac-os-platform). Should we use that instead of the `MACOSX_DEPLOYMENT_TARGET` environment variable?

jcfr: > I didn't know about platform.mac_ver().  Should we use that instead of the MACOSX_DEPLOYMENT_TARGET environment variable?

It is already used to get the target. Your proposed patch makes sense.

> That assumes that MACOSX_DEPLOYMENT_TARGET is always defined on macOS.

You are right, MACOSX_DEPLOYMENT_TARGET is not always defined.


Here is an updated comment:

```
        # If the MACOSX_DEPLOYMENT_TARGET environment variable is defined, use
        # it, as it will be the most accurate. Otherwise use the value returned
        # by platform.mac_ver() provided by the platform module available in the Python standard library.
        # 
        # Note that on macOS, distutils.util.get_platform() is not used because it returns
        # the macOS version on which Python was built which may be significantly older than the user's 
        # current machine.
```

certik: Ah, my bad, for some reason I thought the `distutils.util.get_platform()` was used. 

I am still confused though: one would think the `platform.mac_ver()` would give the correct number, and yet we found that the number in `MACOSX_DEPLOYMENT_TARGET` was correct. How is that possible?

jcfr:
> Would you mind helping me with the Documentation? Where do you think this should be mentioned?

In the `Customizing Deployment Target and Architecture`, I could do the following:

```
Customizing Deployment Target and Architecture
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
.. versionadded:: 0.11.0

Deployment target can be customized by setting the MACOSX_DEPLOYMENT_TARGET
environment variable.


.. versionadded:: 0.7.0

Deployment target and architecture can be customized by associating the 
[...]
```


> Also, how would you suggest to test this? It's not quite clear to me from the test you mentioned.

This could be done changing:

```
@pytest.mark.parametrize("cli_setup_args,keyword_cmake_args,cli_cmake_args,expected_cmake_osx_deployment_target",
```

to 

```
@pytest.mark.parametrize("osx_deployment_target_env_var, cli_setup_args,keyword_cmake_args,cli_cmake_args,expected_cmake_osx_deployment_target",
```

`osx_deployment_target_env_var` should be set to None in all existing case, and a new case setting the env var to a specific value should be added.

CHANGES.rst
```diff
@@ -29,6 +29,9 @@ Bug fixes
 
   * Fix warning using `EXT_SUFFIX` config variable instead of deprecated `SO` variable. See :issue:`381`.
 
+* Honor the `MACOSX_DEPLOYMENT_TARGET` environment variable if it is defined on
+  macOS. See :issue`441`.
```
jcfr:
```
[...] Thanks :user:`certik` for the contribution. See :issue`441`.
```

certik: Fixed in eec72d4815fd6d9fd8ce56f9ca196379a482165d.

skbuild/constants.py
certik: Comment updated in d92bce27dd4dd12c4675e346f5b21c40555ea4b2.

certik:
This is mostly implemented. There are two issues:

* The test fails, but I don't know what the best way to fix it is. I currently set the environment variable in the test, perhaps it does not work somehow.

* There is also an issue that I couldn't figure out how to make the code `flake8` compliant, so I disabled `flake8` testing for now in 8fc4042c72754634ef587bbc18a3af29a70c1044 (that commit should be reverted).

@jcfr if you know how to finish this, feel free to push into my branch. I think it's pretty close, and somebody more familiar with `scikit-build` might be able to finish it quickly.

certik:
The issue is with the test. It is made to execute on all platforms, but the `_default_skbuild_plat_name` does not pick up the `sys.platform == 'darwin'`, and so it will not read the `MACOSX_DEPLOYMENT_TARGET` env var, and so the test will fail on Linux.

certik:
The `_default_skbuild_plat_name()` function is being executed before the test in `tests/test_issue342_cmake_osx_args_in_setup.py` even has a chance to run. And so the function `_default_skbuild_plat_name` cannot be tested like you proposed I think.

Is there another way you would like to test it?

certik:
Tests should pass. @jcfr this is ready for review. The commits should probably be squashed.

Issue #434, opened 2019-06-26
Where are extension modules supposed to be installed?


eric-wieser:
I'm following the example for a simple extension module, but it's installing in a weird place.

When I use:
```cmake
# simulator
add_library(simulator MODULE
 src/simulatormodule.c
)
python_extension_module(simulator)
install(TARGETS simulator LIBRARY DESTINATION lib)
```

And run `pip3.5 install . --user`, I end up with a file at `~/.local/lib/simulator.cpython-35m-x86_64-linux-gnu.so`. I'd like it to end up at `~/.local/python3.5/site-packages/my_package/simulator.cpython-35m-x86_64-linux-gnu.so`. I believe this is already how things work for `distutils.Extension` objects?

Is there an easy way to make this happen? Could the docs be updated to show this type of usage?

thewtex:
Hello @eric-wieser ,

Does changing the `install` `DESTINATION` from `lib` to `my_package` achieve the desire behavior.

The [scikit-build-sample-projects](https://github.com/scikit-build/scikit-build-sample-projects/blob/e38da6d401dae01e4f2467c7106bff405b18410b/projects/hello-cpp/CMakeLists.txt#L9) may also be helpful.

jcfr:
@eric-wieser Thanks for reaching out. Install destination of a module is expected to corresponds to the python package where it is expected to be imported from.

jcfr:
Do you need any other help to move forward ?

eric-wieser:
> Does changing the `install` `DESTINATION` from `lib` to `my_package` achieve the desire behavior.

Yes, that did solve the problem. Can you explain why `lib` resolves to `~/.local/lib/` but `my_package` resolves to `~/.local/python3.5/site-packages/my_package`?

I'm unblocked, but still a little confused about what happened here.

jcfr:
Since there are no python package called 'lib' specified in setup.py, the files to install that are not associated with a package  and are considered as data files and installed there. See https://docs.python.org/3/distutils/setupscript.html#installing-additional-files

eric-wieser:
Ah, that makes sense - I thought the path looked familiar, that is where I remember `data_files` ending up.

Could you point me to the source within scikit-build that makes that decision? It would be great to have it somewhere within the `scikit-build` docs, if it's not already.

jcfr:
> It would be great to have it somewhere within the scikit-build docs, if it's not already.

:+1: (This is tracked by #436)

This happen here:

[...]

PR #442, opened 2019-07-29
[WIP] Add ctypes examle

Erotemic:
I'm having issues similar to #434 which have to do with how scikit-build is finding extension modules. For reference the repo where I'm having these issues is: https://github.com/Erotemic/hesaff/tree/dev/wheels which uses `ctypes` for python bindings.

I also noticed that there was no ctypes example in the samples, so I made one as a MWE of my issue. Once my problem is resolved this MWE should be useful as a sample in the master branch of scikit-build, hence why this is a PR. 

This original library I'm having issues with was based on a pure C++ project and the python bindings were added afterwards, so the CMakeLists is completely standalone, which makes a somewhat different than other scikit-build examples I've seen. 

Because the CMakeLists is builds a pure C++ project, I explicitly point package_data in the `setup.py` file at a pattern that should match the compiled libraries. Ideally running `setup.py` will run CMake, compile the libs, copy them into the correct install/wheel/inplace destination depending on the setup.py command. 

------ 
I have things mostly working and building, the issue is that package data on windows and linux are inconsistent. You'll notice this PR has added some debug statements to `setuptools_wrap.py`. This allows me to determine how scikit-build handles the output of CMake. 

On Linux it does what I expect it to do: 

```
cd $HOME/code/scikit-build/tests/samples/hello-ctypes
python setup.py build_ext --inplace
```

I see
```
-- CLASSIFY FILE --
 * package_data = {'hello': ['*.so']}
 * path = 'hello/libhello.so'
 * found_package = True
```
which means it found `libhello.so` and recognized it as package data, just as I would expect based on my `CMakeLists.txt` and `setup.py`.

However, on Windows:
```
cd %HOME%/code/scikit-build/tests/samples/hello-ctypes
call "C:/Program Files (x86)/Microsoft Visual Studio/2017/BuildTools/VC/Auxiliary/Build/vcvars64.bat"
python setup.py build_ext --inplace -- -G "Visual Studio 15 2017 Win64"
```

The library is put into `<INSTALL>/bin/hello.dll` instead of `<INSTALL>/hello/hello.dll`, which seems to prevent it from registering as a library file, and thus its put into the data folder.
```
-- CLASSIFY FILE --                                                                                                             
 * package_data = {'hello': ['*.dll']}                                                                                          
 * path = 'lib/hello.lib'                                                                                                       
 * fallback to generic data file                                                                                                
-- CLASSIFY FILE --                                                                                                             
 * package_data = {'hello': ['*.dll']}                                                                                          
 * path = 'bin/hello.dll'                                                                                                       
 * fallback to generic data file                                                                                                
```

-----

PS: 

I was about to post this, but I went to double check my MWE against hello-cpp, and I noticed the `add_library` command in the skbuild example i using MODULE instead of SHARED. When I change to use MODULE then CMake installs to `<INSTALL>/hello/hello.dll` and thus skbuild handles the dll appropriately. 

I'm leaving the above post as-is because at the very least, its still think its a gotcha that should be documented. On the other hand does it make sense (ie does scikit-build explicitly require) that your binary-extensions are MODULES and not SHARED libraries? I can think of some instances where you may want to link against a CFFI-ctypes library, so perhaps the inconsistency between windows and linux for SHARED objects is still a concern?

This is getting long so perhaps its best to summarize my questions: 

- [ ] Has scikit-build been designed to CFFI ctypes extensions similar to the example in the PR? Am I doing something this project simply hasn't considered yet? 
- [ ] Am I doing anything blatantly wrong or using bad practices in my example? (I am not a C expert, so I have doubts). 
- [ ] Should `add_library` calls for CFFI libraries always use MODULE or should SHARED also be allowed?

### DIFF
--- a/skbuild/setuptools_wrap.py
+++ b/skbuild/setuptools_wrap.py
@@ -733,6 +733,8 @@ def _classify_installed_files(install_paths, package_data, package_prefixes,
                               scripts, new_scripts,
                               data_files,
                               cmake_source_dir, cmake_install_dir):
+
+    print('-- CLASSIFY INSTALLED FILES --')
     assert not os.path.isabs(cmake_source_dir)
     assert cmake_source_dir != "."
 
@@ -754,12 +756,16 @@ def _classify_installed_files(install_paths, package_data, package_prefixes,
                        py_modules, new_py_modules,
                        scripts, new_scripts,
                        data_files)
+    print('-- DONE CLASSIFY INSTALLED FILES --')
 
 
 def _classify_file(path, package_data, package_prefixes,
                    py_modules, new_py_modules,
                    scripts, new_scripts,
                    data_files):
+    print('-- CLASSIFY FILE --')
+    print(' * package_data = {!r}'.format(package_data))
+    print(' * path = {!r}'.format(path))
     found_package = False
     found_module = False
     found_script = False
@@ -780,6 +786,7 @@ def _classify_file(path, package_data, package_prefixes,
             break
 
     if found_package:
+        print(' * found_package = {!r}'.format(found_package))
         return
     # If control reaches this point, then this installed file is not part of
     # a package.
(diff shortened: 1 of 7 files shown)


henryiii:

PR #443, opened 2019-08-06
Print configuration CMake command line

native-api:
We are using `scikit-build` in https://github.com/skvark/opencv-python. When something goes wrong, we need to run the exact same CMake invocation with tracing to diagnose. This PR adds printing of it.

Instead of printing the raw argv, [`shlex.join`](https://docs.python.org/3.8/library/shlex.html?highlight=shlex#shlex.join) could be used to produce a copypasteable cmdline. But it's 3.8+

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,6 +20,9 @@ New Features
 * Add support for OS/400 (now known as IBM i).
   Thanks :user:`jwoehr` for the contribution. See :issue:`444`.
 
+* Display CMake command used to configure the project.
+  Thanks :user:`native-api` for the contribution. See :issue:`443`.
+
 * CMake modules:
 
   * Improve CMake module :doc:`/cmake-modules/F2PY` adding `add_f2py_target()` CMake function
--- a/skbuild/cmaker.py
+++ b/skbuild/cmaker.py
@@ -2,6 +2,7 @@
 This module provides an interface for invoking CMake executable.
 """
 
+from __future__ import print_function
 import argparse
 import distutils.sysconfig as du_sysconfig
 import glob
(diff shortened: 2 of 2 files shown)


jcfr:
Thanks for the contribution :pray: 

> Instead of printing the raw argv, shlex.join could be used to produce a copypasteable cmdline

Good idea, I improved the existing function `def _formatArgsForDisplay` to display a properly formatted command.

Here is an example output:

```
[...]

Configuring Project
  Working directory:
    /home/jcfr/Projects/scikit-build-sample-projects/projects/hello-cpp/_skbuild/linux-x86_64-3.5/cmake-build
  Command:
    cmake /home/jcfr/Projects/scikit-build-sample-projects/projects/hello-cpp -G Ninja -DCMAKE_INSTALL_PREFIX:PATH=/home/jcfr/Projects/scikit-build-sample-projects/projects/hello-cpp/_skbuild/linux-x86_64-3.5/cmake-install -DPYTHON_EXECUTABLE:FILEPATH=/home/jcfr/.virtualenvs/scikit-build-35/bin/python -DPYTHON_VERSION_STRING:STRING=3.5.0+ -DPYTHON_INCLUDE_DIR:PATH=/usr/include/python3.5m -DPYTHON_LIBRARY:FILEPATH=/usr/lib/x86_64-linux-gnu/libpython3.5m.so -DSKBUILD:BOOL=TRUE -DCMAKE_MODULE_PATH:PATH=/home/jcfr/Projects/scikit-build/skbuild/resources/cmake -DCMAKE_BUILD_TYPE:STRING=Release

[...]
```

I will integrate once the CI completes. Thanks for your patience

PR #448, opened 2019-09-14
[WIP] Add Logging

Erotemic:
The way scikit-build determines which outputs of CMake go where can be confusing. Adding some basic logging to explicitly say what is happening can go a long way in helping to mitigate this confusion. 

I'm still working out all of the details, so this is a WIP for now, while I'm also working on integrating skbuild into FLANN: https://github.com/mariusmuja/flann/pull/425


henryiii:

PR #453, opened 2019-10-03
XDoctest + CMaker doctests

Erotemic:
I'm attempting to better understand this library, so I can get it to work with FLANN in https://github.com/mariusmuja/flann/pull/425 

Writing doctests is an effective way for me to do that because it lets me step through the code in IPython. They're also great for unit tests and helping newcomers see examples of how a library can be used. They can setup dummydata entrypoints into almost any place in your system, which is extremely useful for debugging.

Unfortunately, the builtin Python doctest module isn't that great when it comes to parsing and running doctests on a CI, but that's ok because I've written [xdoctest](https://github.com/Erotemic/xdoctest), which gets rid of a lot of the weirdness and fundamental restrictions present in the builtin module. 

The doctests are pretty simple. The one for CMaker is a near-copy of a unit test, but it doesn't rely on pytest fixtures, so it can be copy-pasted and hence is much more suitable as an example (I find pytest fixtures frustrating because it is not easy to create instances of them outside of the pytest framework). 

Hopefully someone gets some time to look at this. The scope and diff is pretty small. Even if this isn't merged right away, it would be useful to know if this would be accepted.


thewtex:
@Erotemic very nice!! LGTM. @jcfr what do you think?

Erotemic:
@thewtex rebased to resolve conflicts with current master branch.

henryiii:
More tests and docs is always a good thing. Thanks! :)

PR #458, opened 2019-11-24
Replace linux_distribution() missing in Python 3.8

bnavigator:
`platform.linux_distribution()` was removed in Python 3.8. Move to `distro.id()`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 wheel>=0.29.0
 setuptools>=28.0.0
 packaging
+distro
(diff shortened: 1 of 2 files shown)


jcfr:
@bnavigator Thanks for the contribution
@thewtex Thanks for reviewing and integrating

:pray:

jcfr:
@bnavigator I suggest you submit an other PR adding an entry to https://github.com/scikit-build/scikit-build/blob/master/CHANGES.rst

PR #463, opened 2019-12-27
Fix bug in python.h lookup logic.

aws-taylor:
Looking at the diff of https://github.com/scikit-build/scikit-build/commit/1c11225560cdefff592cdfd71c9d4f3777765606 it appears that DeMorgan's law bit us here. The net result is that scikit build works if the location of python.h pointed to by sysconfig is accurate, but in a cross-compile type scenario Python.h may be incorrectly marked as found.

### DIFF
--- a/skbuild/cmaker.py
+++ b/skbuild/cmaker.py
@@ -268,7 +268,7 @@ class CMaker(object):
         # if Python.h not found (or python_include_dir is None), try to find a
         # suitable include dir
         found_python_h = (
-            python_include_dir is not None or
+            python_include_dir is not None and
             os.path.exists(os.path.join(python_include_dir, 'Python.h'))
         )
 
--- a/skbuild/cmaker.py
+++ b/skbuild/cmaker.py
@@ -322,6 +322,15 @@ class CMaker(object):
             if python_version:
                 candidate_versions += ('',)
 
+                pymalloc = None
+                try:
+                    pymalloc = bool(sysconfig.get_config_var('WITH_PYMALLOC'))
+                except AttributeError:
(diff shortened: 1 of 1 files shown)


dralley:
@thewtex @aws-taylor Is this related to a problem I've been experiencing when trying to build wheels against ```manylinux2014```?

```
  Linking C shared library libcomps/_libpycomps.so
  /opt/rh/devtoolset-8/root/usr/libexec/gcc/x86_64-redhat-linux/8/ld: cannot find -lpython3.6m
```

https://github.com/dralley/libcomps/runs/447321284?check_suite_focus=true#step:7:307

If not, how can I tell whether this issue is related to ```scikit-build``` or some other cause?

FWIW, I've tested manually building the wheel in Centos 7, so something about the ```manylinux2014``` environment is causing the difference.

thewtex:
>   /opt/rh/devtoolset-8/root/usr/libexec/gcc/x86_64-redhat-linux/8/ld: cannot find -lpython3.6m

This is a library linking error versus an include directory error.

From those messages, it is difficult to know the cause of the issue. However, the next steps are to check for Python related variables in the build's *CMakeCache.txt* file. If the Python library was located correctly, the full path will be identified. On the other hand, if building wheels for Linux, [weak linking should be used](https://scikit-build.readthedocs.io/en/latest/cmake-modules/targetLinkLibrariesWithDynamicLookup.html), i.e. the build should not be linking to the Python library at all.

dralley:
@thewtex I did figure it out eventually; I forgot I made this comment.  Sorry!  But, thanks for answering anyway

PR #464, opened 2019-12-27
Append 'm' suffix to python version if compiled with pymalloc.

aws-taylor:
If python is built with pymalloc, then an 'm' suffix is appended to the version. For example, in a 3.6 pymalloc build, Python.h would live under 'include/python3.6m/Python.h', breaking the current lookup logic.

### DIFF
--- a/skbuild/cmaker.py
+++ b/skbuild/cmaker.py
@@ -322,6 +322,15 @@ class CMaker(object):
             if python_version:
                 candidate_versions += ('',)
 
+                pymalloc = None
+                try:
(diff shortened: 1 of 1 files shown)


thewtex:
Added in #463

karlsonbest:
Hi All,
Have question, How I can get info about this m suffix from command line ? 
for example sys.version_info 
now output is sys.version_info(major=3, minor=7, micro=8, releaselevel='final', serial=0)
would be great --> sys.version_info(major=3, minor=7**m**, micro=8, releaselevel='final', serial=0) or additional attribute

PR #465, opened 2019-12-27
Add stdlib and platlib to the list of directories searched when looki…

aws-taylor:
…ng for python libraries.

This mirrors the logic already in place for searching for include directories. Prior to this change, only sysconfig was used to find python libraries, however in a cross-compile situation the location in which python was built may not match its current location and thus library lookup would fail.


thewtex:
CI is failing

henryiii:

PR #475, opened 2020-03-30
Ignore DESTDIR when installing to _skbuild

jokva:
The cmake --build --target install command is used to move built
artefacts to _skbuild/, but when used in an environment with staged
installs (DESTDIR [1] [2]) the _skbuild directory also ends up there.
This is never desired as the _skbuild is an internal detail to the
scikit-build process, and DESTDIR is not meant to be used to specify the
build location.

[1] Happens in larger projects where scikit-build is not the only player
[2] https://github.com/equinor/segyio/blob/0c791d0f0ae4973058a51e2014837946e668902b/python/setup.py

### DIFF
--- a/skbuild/cmaker.py
+++ b/skbuild/cmaker.py
@@ -23,6 +23,7 @@ from .constants import (CMAKE_BUILD_DIR,
                         SETUPTOOLS_INSTALL_DIR)
 from .platform_specifics import get_platform
 from .exceptions import SKBuildError
+from .utils import push_env
 
 RE_FILE_INSTALL = re.compile(
     r"""[ \t]*file\(INSTALL DESTINATION "([^"]+)".*"([^"]+)"\).*""")
@@ -85,6 +86,32 @@ def get_cmake_version(cmake_executable=CMAKE_DEFAULT_EXECUTABLE):
         version_string = version_string.decode()
 
     return version_string.splitlines()[0].split(' ')[-1]
+
+
+class push_env_without_destdir(push_env):
+    """Remove DESTDIR from MAKEFLAGS and environment
+
+    When scikit-build is invoked as a part of a larger cmake project, makefile,
+    or something else that sets the DESTDIR environment variable, the internal
+    cmake --target install honours it and makes artefacts under DESTDIR/
+
+    The internal install into _skbuild/ should never need to respect DESTDIR,
+    so simply remove it from the environment. When cmake generates the
+    makefiles, it stores the make parameters [1] in the MAKEFLAGS environment
+    variable.  The DESTDIR entry needs to be removed from here as well.
+
+    [1] make DESTDIR=/path install
+    """
+    def __init__(self, **kwargs):
+        kwargs['DESTDIR'] = None
+
+        if 'MAKEFLAGS' in kwargs:
+            makeflags = kwargs['MAKEFLAGS']
+            flags = makeflags.split(' ')
+            flags = [flag for flag in flags if not flag.startswith('DESTDIR=')]
+            kwargs['MAKEFLAGS'] = ' '.join(flags)
+
+        super(push_env_without_destdir, self).__init__(**kwargs)
 
 
 class CMaker(object):
@@ -491,7 +518,15 @@ class CMaker(object):
                    shlex.split(os.environ.get("SKBUILD_BUILD_OPTIONS", "")))
         )
 
-        rtn = subprocess.call(cmd, cwd=CMAKE_BUILD_DIR(), env=env)
+        # The subprocess.call() expects a modified environment or None (means
+        # inherit process, i.e. os.environ). push_env_without_destdir, however,
+        # uses **kwargs to set/unset parameters
+        if env is None:
+            env = {}
+
+        with push_env_without_destdir(**env):
+            rtn = subprocess.call(cmd, cwd=CMAKE_BUILD_DIR(), env=None)
+
         if rtn != 0:
             raise SKBuildError(
                 "An error occurred while building with CMake.\n"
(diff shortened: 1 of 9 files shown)


jokva:
Alright, this should be more-or-less complete now. I'm not sure about the failing tests, they shouldn't be affected at all by this patch.

skbuild/cmaker.py
```diff
@@ -491,6 +519,8 @@ def make(self, clargs=(), config="Release", source_dir=".", env=None):
                    shlex.split(os.environ.get("SKBUILD_BUILD_OPTIONS", "")))
         )
 
+        env = env_without_destdir(env)
```
jcfr: Instead, you would then update

```
def make(self, clargs=(), config="Release", source_dir=".", env=None):
```

to be something like this

```
@env_without_destdir
def make(self, clargs=(), config="Release", source_dir=".", env=None):
```

jcfr (changes requested):

skbuild/cmaker.py
```diff
@@ -87,6 +87,34 @@ def get_cmake_version(cmake_executable=CMAKE_DEFAULT_EXECUTABLE):
     return version_string.splitlines()[0].split(' ')[-1]
 
 
+def env_without_destdir(env):
```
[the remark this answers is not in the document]
jcfr: Last, would be stellar to update the restructured text documentation to document this behavior.

This could be done adding a section `Environment` before the `Advanced Usage` section in [docs/usage.rst#advanced-usage](https://github.com/scikit-build/scikit-build/blob/master/docs/usage.rst#advanced-usage)

For now the section would simply list each variables and the effect of scikit-build on each:

```
* DESTDIR: ...

* MAKEFLAGS: ...
```

skbuild/cmaker.py
jokva: I don't think this will work out well, at least not without breaking the interface. The env= parameter is forwarded to subprocess.call, which expects either None (inherit parent's environment), or a specific environment as a dict. Unless the context decorator implements dict as well, some translation is needed.

Also, it would have to quietly hijack the env= argument. I ended up writing it as a context manager though, but with an explicit with. One benefit I think is the communication of intent, it's really _just_ for the internal install target the environment variables should be ignored.

skbuild/cmaker.py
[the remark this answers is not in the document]
jokva: Done.

skbuild/cmaker.py
```diff
@@ -87,6 +88,32 @@ def get_cmake_version(cmake_executable=CMAKE_DEFAULT_EXECUTABLE):
...
+    or something else that sets the DESTDIR environment variable, the internal
+    cmake --target install honours it and makes artefacts under DESTDIR/
+
+    The internal install into _skbuild/ should never need to respect DESTDIR,
+    so simply remove it from the environment. When cmake generates the
+    makefiles, it stores the make parameters [1] in the MAKEFLAGS environment
+    variable.  The DESTDIR entry needs to be removed from here as well.
+
+    [1] make DESTDIR=/path install
+    """
```
jcfr: What do you think of removing the constructor argument and doing the following:

```python
    def __init__(self):
        makeflags = os.environ.get('MAKEFLAGS', None)

        if makeflags is not None:
            flags = [flag for flag in makeflags.split(' ') if not flag.startswith('DESTDIR=')]
            makeflags = ' '.join(flags)

        super(push_env_without_destdir, self).__init__(DESTDIR=None, MAKEFLAGS=makeflags)
```

skbuild/cmaker.py
```diff
@@ -491,7 +518,15 @@ def make(self, clargs=(), config="Release", source_dir=".", env=None):
...
-        rtn = subprocess.call(cmd, cwd=CMAKE_BUILD_DIR(), env=env)
+        # The subprocess.call() expects a modified environment or None (means
+        # inherit process, i.e. os.environ). push_env_without_destdir, however,
+        # uses **kwargs to set/unset parameters
+        if env is None:
+            env = {}
+
+        with push_env_without_destdir(**env):
+            rtn = subprocess.call(cmd, cwd=CMAKE_BUILD_DIR(), env=None)
+
```
jcfr: Then, we could do this:

```
      with push_env_without_destdir():
            rtn = subprocess.call(cmd, cwd=CMAKE_BUILD_DIR(), env=os.environ)
```

jcfr (commented):
Thanks :pray: for the quick turnaround.

skbuild/cmaker.py
jokva: It still wouldn't all callers of cmaker.make() to pass a specific env. I'm not aware of this feature being _used_ at all (if it is not, why is the env argument bubbled out?). If the env argument to make() can go away, so do all of these problems and the context manager will become a lot simpler.

jcfr: Good point ... I missed that. Sorry for the noise.

The env parameter is indeed important and your implementation makes sense.

https://github.com/scikit-build/scikit-build/blob/0c33f15b0c0452e8bd70ab6057b4d283d6025b51/skbuild/setuptools_wrap.py#L579-L589

jcfr:
Waiting the CI issues are fixed, I will manually test on windows and then integrate. Thanks for your patience.

jokva:
Not a problem :---) I've applied the same hack in my downstream problems, which I can revert when this is integrated (and released) here.

hobu:
Just to chime in that I ran into this and it was very confusing. If this can't be merged, a simple warning shouting to the user that DESTDIR is set would hopefully give them the hint.

henryiii:

PR #476, opened 2020-04-07
Add hook to process manifest

### LINKED ISSUE #473: Cleanup wheels from useless artefacts
Is it possible to exclude  items which are in the cmake install manifest from the wheel?

As it turns out, we were able to do a fully static build of everything in our scikit-build wheel to avoid relocation issues, but te resulting wheel includes the `.a`, as well as the cmake config files and other artefacts from the dependencies. Since all we produce is an executable, we could completely remove all of that...

SylvainCorlay:
This is a simple hook to enable custom processing of the cmake manifest before producing the wheel.

The goal for us is to drop static binaries of dependencies, as well as their headers, and other artefacts such as cmake files, as we want the wheel to only include what is required at runtime.

I am using this here: https://github.com/jupyter-xeus/xeus-python-wheel/pull/36

Fixes #473

### DIFF
--- a/skbuild/setuptools_wrap.py
+++ b/skbuild/setuptools_wrap.py
@@ -609,7 +609,13 @@ def setup(*args, **kw):  # noqa: C901
 
     package_prefixes = _collect_package_prefixes(package_dir, packages)
 
-    _classify_installed_files(cmkr.install(), package_data, package_prefixes,
+    # This hook enables custom processing of the cmake manifest
+    cmake_manifest = cmkr.install()
+    process_manifest = kw.get('cmake_process_manifest_hook')
+    if callable(process_manifest):
+        cmake_manifest = process_manifest(cmake_manifest)
+
+    _classify_installed_files(cmake_manifest, package_data, package_prefixes,
                               py_modules, new_py_modules,
                               scripts, new_scripts,
                               data_files,
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -40,6 +40,8 @@ New Features
 Bug fixes
 ---------
 
+* Fix python 2.7 installation ensuring setuptools < 45 is required. See :issue:`478`.
+
 * Fix unclosed file resource in :meth:`skbuild.cmaker.CMaker.check_for_bad_installs`.
   Thanks :user:`Nic30` for the suggestion. See :issue:`429`.
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,7 @@ New Features
 * Add a hook to process the cmake install manifest building the wheel. The hook
   function can be specified as an argument to the `setup()` function. This can be used e.g.
   to prevent installing cmake configuration files, headers, or static libraries with the wheel.
+  Thanks :user:`SylvainCorlay` for the contribution. See :issue:`473`.
 
 * Add support for passing :ref:`CMake configure options <usage_cmake_configure_options>` like `-DFOO:STRING:bar`
   as global `setuptools` or `pip` options.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,10 @@ Next Release
 New Features
 ------------
 
+* Add a hook to process the cmake install manifest building the wheel. The hook
+  function can be specified as an argument to the `setup()` function. This can be used e.g.
+  to prevent installing cmake configuration files, headers, or static libraries with the wheel.
+
 * Add support for passing :ref:`CMake configure options <usage_cmake_configure_options>` like `-DFOO:STRING:bar`
   as global `setuptools` or `pip` options.
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -78,6 +78,9 @@ Tests
   * Update Azure Pipelines configuration for running tests using `PyPy3 <https://pypy.org/>`_ on Linux.
     Thanks :user:`mattip` for the contribution. See :issue:`418`.
 
+* Update :func:`initialize_git_repo_and_commit` to prevent signing message on system with commit signing
+  enabled globally.
+
 Scikit-build 0.10.0
 ===================
 
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -140,6 +140,21 @@ For example::
 - ``cmake_source_dir``: Relative directory containing the project ``CMakeLists.txt``.
   By default, it is set to the top-level directory where ``setup.py`` is found.
 
+- ``cmake_manifest_process_hook`: Python function consumming the list of files to be
+  installed produced by cmake. For example, `cmake_manifest_process_hook` can be used
+  to exclude static libraries from the built wheel.
+
+For example::
+
+    def exclude_static_libraries(cmake_manifest):
+        return list(filter(lambda name: not (name.endswith('.a')), cmake_manifest))
+
+    setup(
+      [...]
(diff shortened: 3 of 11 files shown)


SylvainCorlay:
It seems that the Python 2.7 builds need to be disabled as the installed version of setuptools requires Python>=3.5

jcfr:
Thanks for the contribution 🙏 

I will be able to review tomorrow.

SylvainCorlay:
> I will be able to review tomorrow.

Thanks!  If you are OK with it, I am happy to add a bit of documentation in the relevant section. I was able to use it successfully in https://github.com/jupyter-xeus/xeus-python-wheel/pull/36.

If there is a release of scikit-build with that hook, that would be great.

thewtex:
@jcfr why are the CI tests failing?

jcfr:
>  why are the CI tests failing?

It is because the latest version of setuptools is not working with 2.7 ... I got distracted on Friday and didn't move forward with publishing a fix.

Stay tuned for an update. :pray:

SylvainCorlay:
@jcfr I am happy to rebase this PR if you update master.

SylvainCorlay:
If someone has the time to give this a look, it would be super appreciated.

jcfr:
@SylvainCorlay I didn't forget ... it is definitely in the queue. Thanks for your patience :pray:

skbuild/setuptools_wrap.py
```diff
@@ -609,7 +609,13 @@ def setup(*args, **kw):  # noqa: C901
 
     package_prefixes = _collect_package_prefixes(package_dir, packages)
 
-    _classify_installed_files(cmkr.install(), package_data, package_prefixes,
+    # This hook enables custom processing of the cmake manifest
+    cmake_manifest = cmkr.install()
+    process_manifest = kw.get('process_manifest')
+    if process_manifest is not None:
```
jcfr, line 615: May be worth checking that `process_manifest` is callable

skbuild/setuptools_wrap.py
```diff
@@ -609,7 +609,13 @@ def setup(*args, **kw):  # noqa: C901
 
     package_prefixes = _collect_package_prefixes(package_dir, packages)
 
-    _classify_installed_files(cmkr.install(), package_data, package_prefixes,
+    # This hook enables custom processing of the cmake manifest
+    cmake_manifest = cmkr.install()
+    process_manifest = kw.get('process_manifest')
```
jcfr, line 614: Since all skbuild options starts with `cmake_` (see [here](https://scikit-build.readthedocs.io/en/latest/usage.html#scikit-build-options)),  what about using the more explicit name `cmake_process_manifest_hook` ?

SylvainCorlay: Fixed. Now naming the argument `cmake_process_manifest_hook`.

skbuild/setuptools_wrap.py
SylvainCorlay: Done, now checking that `process_manifest` is callable.

SylvainCorlay:
> Before we move forward, could you:
> - update documentation to describe the keyword and in which case it could be used
> - update CHANGES.rst
> - add a simple test

Done! This is ready to o @jcfr

jcfr:
Thanks, I will review later tomorrow or Monday. Have a good weekend.

SylvainCorlay:
Hey, let me know if there is anything else needed. This is on our critical path to publish the xeus-python wheels.

We are currently using scikit-build from this branch, but this is not great for reproducibility.

thewtex:
ping @jcfr

SylvainCorlay:
Hey @jcfr is there anything we can do to help move this forward?

skbuild/setuptools_wrap.py
```diff
@@ -609,7 +609,13 @@ def setup(*args, **kw):  # noqa: C901
 
     package_prefixes = _collect_package_prefixes(package_dir, packages)
 
-    _classify_installed_files(cmkr.install(), package_data, package_prefixes,
+    # This hook enables custom processing of the cmake manifest
+    cmake_manifest = cmkr.install()
+    process_manifest = kw.get('cmake_process_manifest_hook')
+    if callable(process_manifest):
```
jcfr, line 615: It may be sensible to report an error if it is not callable

jcfr:
@SylvainCorlay I will add the error reporting and integrate. No further action needed.

jcfr:
The test `tests/test_filter_manifest.py::test_bdist_wheel_command` is failing on macOS (for all python versions) with the following error:

```
lib_file = <_io.BufferedReader name='_skbuild/macosx-10.13-x86_64-3.7/setuptools/bdist.macosx-10.13-x86_64/wheel/hello/_swig_mwe.so'>
seek = 0
    def get_base_class_and_magic_number(lib_file, seek=None):
        if seek is None:
            seek = lib_file.tell()
        else:
            lib_file.seek(seek)
        magic_number = ctypes.c_uint32.from_buffer_copy(
>           lib_file.read(ctypes.sizeof(ctypes.c_uint32))).value
E       ValueError: Buffer size too small (0 instead of at least 4 bytes)
../../../.pyenv/versions/3.7.0/lib/python3.7/site-packages/wheel/macosx_libfile.py:224: ValueError
```

This is caused by the dummy `.so` file being an invalid library. In that case, the following function fail to extract macOS information: https://github.com/pypa/wheel/blob/a51977075740fda01b2c0e983a79bfe753567219/src/wheel/macosx_libfile.py#L219-L225

SylvainCorlay:
Maybe I should include a real compiled extension rather than an empty dummy file for this example - although the other tests that do the same will still be failing.

jcfr:
No worry, I will finalize this morning. Look like I missed something.

jcfr:
CI is mostly green, after Travis completes I will integrate.

[Linux_PyPy3](https://dev.azure.com/scikit-build/scikit-build/_build/results?buildId=284&view=logs&j=b18cf24a-3d78-5967-8f9a-813d25a10c78) failures are unrelated and will be fixed later or even disabled.

SylvainCorlay:
Great! Thanks for taking the time!

jcfr:
> > if callable(process_manifest):
> It may be sensible to report an error if it is not callable

I created an issue to track this. See https://github.com/scikit-build/scikit-build/issues/482

This would be a good first issue for anyone new to the code base

jcfr:
> Great! Thanks for taking the time!

Thanks to you for contributing :pray: 

Have a good weekend

SylvainCorlay:
Thanks for merging!

Would be great to have a patch release with that feature! We could start targetting it in xeus-python-wheel!

SylvainCorlay:
@jcfr it would be fantastic if there was a scikit-build tag including this patch!

We have been looking into streamlining the release of xeus-python on PyPI for a few weeks and having a new release of scikit-build is the main blocker.

jcfr:
Yes, I will do a release shortly. After integrating Pablo PR.

SylvainCorlay:
Great thanks!

SylvainCorlay:
> Yes, I will do a release shortly. After integrating Pablo PR.

Is there any blocker on this ! I don't want to be nagging - but I am following up as we have been blocked on this for a few weeks!

jcfr:
Fundamentally there are no blockers, time is just finite resource.

After https://github.com/scikit-build/scikit-build/pull/486 complete I will do release :tada:

PR #477, opened 2020-04-09
Add cmake_target to change the target used by CMake

phcerdan:
Parses the command line for the `--target` option.
And also provide the `cmake_install_target` parameter in `setup.py`

The command line argument has preference over the `cmake_install_target` from
`setup.py`.

This allows the user to point to targets other than 'install'.
The user has to set a target in CMake to install only a specific
component.

```cmake
install(TARGETS foo COMPONENT runtime)
add_custom_target(foo-install-runtime
    ${CMAKE_COMMAND}
    -DCMAKE_INSTALL_COMPONENT=runtime
    -P "${PROJECT_BINARY_DIR}/cmake_install.cmake")
```

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -175,6 +175,21 @@ For example::
 - ``cmake_minimum_required_version``: String identifying the minimum version of CMake required
   to configure the project.
 
+- ``cmake_install_target``: Name of the target to "build" for installing the artifacts into the wheel.
+  By default, this option is set to ``install``, which is always provided by CMake.
+  This can be used to only install certain components.
+
+For example::
+
+    install(TARGETS foo COMPONENT runtime)
+    add_custom_target(foo-install-runtime
+        ${CMAKE_COMMAND}
+        -DCMAKE_INSTALL_COMPONENT=runtime
+        -P "${PROJECT_BINARY_DIR}/cmake_install.cmake"
+        DEPENDS foo
+        )
+
+
 Scikit-build changes the following options:
 
 .. versionadded:: 0.7.0
--- a/skbuild/cmaker.py
+++ b/skbuild/cmaker.py
@@ -573,17 +573,57 @@ class CMaker(object):
                     ("      " + _install) for _install in bad_installs)
             )))
 
-    def make(self, clargs=(), config="Release", source_dir=".", env=None):
+    def make(self, clargs=(), config="Release", source_dir=".",
+             install_target="install", env=None):
         """Calls the system-specific make program to compile code.
+
+           install_target: string
+                Name of the target responsible to install the project.
+                Default is "install".
+
+                .. note::
+
+                   To workaround CMake issue #8438.
+                   See https://gitlab.kitware.com/cmake/cmake/-/issues/8438
+                   Due to a limitation of CMake preventing from adding a dependency
+                   on the "build-all" built-in target, we explicitly build the project first when
+                   the install target is different from the default on.
         """
         clargs, config = pop_arg('--config', clargs, config)
+        clargs, install_target = pop_arg('--install-target', clargs, install_target)
         if not os.path.exists(CMAKE_BUILD_DIR()):
             raise SKBuildError(("CMake build folder ({}) does not exist. "
                                 "Did you forget to run configure before "
                                 "make?").format(CMAKE_BUILD_DIR()))
 
-        cmd = [self.cmake_executable, "--build", source_dir,
-               "--target", "install", "--config", config, "--"]
+        # Workaround CMake issue #8438
+        # See https://gitlab.kitware.com/cmake/cmake/-/issues/8438
+        # Due to a limitation of CMake preventing from adding a dependency
+        # on the "build-all" built-in target, we explicitly build
+        # the project first when
+        # the install target is different from the default on.
+        if install_target != "install":
+            self.make_impl(clargs=clargs, config=config, source_dir=source_dir,
+                           install_target=None, env=env)
+
+        self.make_impl(clargs=clargs, config=config, source_dir=source_dir,
+                       install_target=install_target, env=env)
+
+    def make_impl(self, clargs, config, source_dir, install_target, env=None):
+        """
+        Precondition: clargs does not have --config nor --install-target options.
+        These command line arguments are extracted in the caller function
+        `make` with `clargs, config = pop_arg('--config', clargs, config)`
+
+        This is a refactor effort for calling the function `make` twice in
+        case the install_target is different than the default `install`.
+        """
+        if not install_target:
+            cmd = [self.cmake_executable, "--build", source_dir,
+                   "--config", config, "--"]
+        else:
+            cmd = [self.cmake_executable, "--build", source_dir,
+                   "--target", install_target, "--config", config, "--"]
         cmd.extend(clargs)
         cmd.extend(
             filter(bool,
@@ -591,17 +631,25 @@ class CMaker(object):
         )
 
         rtn = subprocess.call(cmd, cwd=CMAKE_BUILD_DIR(), env=env)
+        # For reporting errors (if any)
+        if not install_target:
+            install_target = "internal build step [valid]"
+
         if rtn != 0:
             raise SKBuildError(
                 "An error occurred while building with CMake.\n"
                 "  Command:\n"
                 "    {}\n"
+                "  Install target:\n"
+                "    {}\n"
                 "  Source directory:\n"
                 "    {}\n"
                 "  Working directory:\n"
                 "    {}\n"
-                "Please see CMake's output for more information.".format(
+                "Please check the install target is valid and see CMake's output for more "
+                "information.".format(
                     self._formatArgsForDisplay(cmd),
+                    install_target,
                     os.path.abspath(source_dir),
                     os.path.abspath(CMAKE_BUILD_DIR())))
 
--- a/skbuild/setuptools_wrap.py
+++ b/skbuild/setuptools_wrap.py
@@ -70,6 +70,11 @@ def create_skbuild_argparser():
         help='specify the path to the cmake executable'
     )
     parser.add_argument(
+        '--install-target', default=None, metavar='',
+        help='specify the CMake target performing the install. '
+             'If not provided, uses the target ``install``'
+    )
+    parser.add_argument(
         '--skip-generator-test', action='store_true',
         help='skip generator test when a generator is explicitly selected using --generator'
     )
@@ -112,6 +117,8 @@ def parse_skbuild_args(args, cmake_args, build_tool_args):
     build_tool_args.extend(['--config', namespace.build_type])
     if namespace.jobs is not None:
         build_tool_args.extend(['-j', str(namespace.jobs)])
+    if namespace.install_target is not None:
+        build_tool_args.extend(['--install-target', namespace.install_target])
 
     if namespace.generator is None and namespace.skip_generator_test is True:
         sys.exit("ERROR: Specifying --skip-generator-test requires --generator to also be specified.")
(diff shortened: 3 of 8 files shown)


jcfr:
Thanks for the contribution 🙏 

I will be able to review tomorrow.

phcerdan:
Not sure about the errors in travis, appveyor and azure, they seem unrelated, but let me know if I can help.

jcfr:
I will rebase the topic after the fix addressing the Continuous Integration errors is merged. Thanks for your patience :pray:

tests/test_cmaker.py
```diff
@@ -130,6 +130,56 @@ def test_make(configure_with_cmake_source_dir, capfd):
             assert message in out
 
 
+@pytest.mark.parametrize("target", ("install", "install-runtime", "bad_value"))
```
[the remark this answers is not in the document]
phcerdan: I added the empty string. And it doesn't generate an error, it will take the default target of the maker.

However not sure if this breaks any future assumptions that `maker` performs an install step.

tests/test_cmaker.py
```diff
@@ -130,6 +130,56 @@ def test_make(configure_with_cmake_source_dir, capfd):
...
+            if target == "install" or target == "install-runtime":
+                cmkr.make(target=target, env=env)
+            else:
+                try:
+                    cmkr.make(target="bad_value", env=env)
+                except SKBuildError:
+                    pass
+
+        out, err = capfd.readouterr()
+        # This message apperas with both targets: default 'install' and
```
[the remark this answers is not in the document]
phcerdan: Done

skbuild/setuptools_wrap.py
```diff
@@ -489,6 +497,20 @@ def setup(*args, **kw):  # noqa: C901
     # one is considered, let's prepend the one provided in the setup call.
     cmake_args = skbuild_kw['cmake_args'] + cmake_args
 
+    # Handle cmake_target
+    # get the target (next item after '--target') or return '' if no --target
+    cmake_target_from_command = next(
+        (make_args[index+1] for index, item in enumerate(make_args) if
+         item == '--target'), '')
+    cmake_target_from_setup = skbuild_kw['cmake_target']
+    # Setting target form command takes preference
```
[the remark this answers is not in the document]
phcerdan: Done

tests/test_cmaker.py
```diff
@@ -130,6 +130,56 @@ def test_make(configure_with_cmake_source_dir, capfd):
             assert message in out
 
 
+@pytest.mark.parametrize("target", ("install", "install-runtime", "bad_value"))
+def test_make_with_target(target, capfd):
+    tmp_dir = _tmpdir('test_make_with_target')
+    with push_dir(str(tmp_dir)):
+
+        tmp_dir.join('CMakeLists.txt').write(textwrap.dedent(
```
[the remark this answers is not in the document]
phcerdan: Maybe, but not pursued.

tests/test_cmaker.py
```diff
@@ -130,6 +130,56 @@ def test_make(configure_with_cmake_source_dir, capfd):
...
+              -DCMAKE_INSTALL_COMPONENT=runtime
+              -P "${PROJECT_BINARY_DIR}/cmake_install.cmake"
+              )
+            """
+        ))
+
+        with push_dir(str(tmp_dir)):
+            cmkr = CMaker()
+            env = cmkr.configure()
+            if target == "install" or target == "install-runtime":
```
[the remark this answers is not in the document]
phcerdan: Done

tests/test_cmaker.py
```diff
@@ -130,6 +130,56 @@ def test_make(configure_with_cmake_source_dir, capfd):
...
+                cmkr.make(target=target, env=env)
+            else:
```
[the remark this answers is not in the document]
phcerdan: Done

phcerdan:
I guess this option is another way to solve #473. Also handled via MANIFEST files in: #476

docs/usage.rst
```diff
@@ -175,6 +175,21 @@ For example::
 - ``cmake_minimum_required_version``: String identifying the minimum version of CMake required
   to configure the project.
 
+- ``cmake_target``: Target generated by CMake that will be used in the compilation step.
```
jcfr, line 178: What do you think of renaming this to `cmake_install_target` ? 

Suggestion:

```
Target generated by CMake that will be used in the compilation step.
By default, this option is set to ``install``, which is always provided by CMake.
This can be used to only install certain components.
```

->

```
Name of the target to "build" for installing the artifacts into the wheel.
By default, this option is set to ``install``, which is always provided by CMake.
This can be used to only install certain components.
```

phcerdan: I like it. But not sure if you foresee any utility on using a non-install target in the future. Your call!

phcerdan: Thinking a little bit more. The `install` target depends on the `build-all` internal target. So, you will never do a `make install` or `ninja install` without triggering the build-all.

However, with custom targets, like `install-runtime`, the build is not guaranteed. The user should make that new install target depends on preexisting targets to force the build. It would a nice CMake feature to expose the `build-all` target to alleviate this [issue](https://gitlab.kitware.com/cmake/cmake/-/issues/8438).

So, thinking twice, I don't think cmake_install_target is that great. Because, the user should take care of handling the dependencies of that new install target... But I am not sure, as I said, your call.

phcerdan: Suggestion integrated in last amend.

jcfr: Considering that scikit-build explicitly build the install target and there is no way to add a target depending on a built-in CMake targets, the proposed feature makes sense. Thanks for referencing the CMake issue. :100:

jcfr: > the build is not guaranteed

I missed that ... 

Then in case `cmake_install_target` is specified and different from the default value,  should we do this:

```
        cmd = [self.cmake_executable, "--build", source_dir, "--config", config, "--"]
        cmd.extend(clargs)
```

followed by an other call to

```
        cmd = [self.cmake_executable, "--build", source_dir, "--target", target, "--config", config, "--"]
        cmd.extend(clargs)
```

This would ensure the project is always built before building the custom install targets.

phcerdan: That is great! Integrating it...

docs/usage.rst
```diff
@@ -175,6 +175,21 @@ For example::
 - ``cmake_minimum_required_version``: String identifying the minimum version of CMake required
   to configure the project.
 
+- ``cmake_target``: Name of the target to "build" for installing the artifacts into the wheel.
```
jcfr, line 178: What about renaming to `cmake_install_target` ?

skbuild/cmaker.py
```diff
@@ -474,17 +474,42 @@ def check_for_bad_installs():
...
             )))
 
-    def make(self, clargs=(), config="Release", source_dir=".", env=None):
+    def make(self, clargs=(), config="Release", source_dir=".",
+             target="install", env=None):
         """Calls the system-specific make program to compile code.
+
+           target: string
+               Target generated by CMake performing the install.
+               The target ``install`` is the default. It is always
```
jcfr, line 483: `It is always generated ...` -> `It is always added ...`

skbuild/cmaker.py
```diff
@@ -474,17 +474,42 @@ def check_for_bad_installs():
                     ("      " + _install) for _install in bad_installs)
             )))
 
-    def make(self, clargs=(), config="Release", source_dir=".", env=None):
+    def make(self, clargs=(), config="Release", source_dir=".",
+             target="install", env=None):
```
jcfr, line 478: `target` -> `install_target`

skbuild/cmaker.py
```diff
@@ -474,17 +474,42 @@ def check_for_bad_installs():
...
             raise SKBuildError(("CMake build folder ({}) does not exist. "
                                 "Did you forget to run configure before "
                                 "make?").format(CMAKE_BUILD_DIR()))
 
+        # The default target `install` depends on the builtin target: build-all
+        # This ensures that `make install` always trigger a build.
+        # However, this is not guaranteed by other targets, so we build the
+        # project first.
+        # There is a CMake PR to expose the build-all internal target:
+        # https://gitlab.kitware.com/cmake/cmake/-/issues/8438
```
jcfr, line 505:
```
        # The default target `install` depends on the builtin target: build-all
        # This ensures that `make install` always trigger a build.
        # However, this is not guaranteed by other targets, so we build the
        # project first.
        # There is a CMake PR to expose the build-all internal target:
        # https://gitlab.kitware.com/cmake/cmake/-/issues/8438
```

->

```
        # Workaround CMake issue #8438 
        # See https://gitlab.kitware.com/cmake/cmake/-/issues/8438
        # Due to a limitation of CMake preventing from adding a dependency
        # on the "build-all" built-in target, we explicitly build the project first when
        # the install target is different from the default on.
```

jcfr: Consider also moving the regular build (without the install target) before the actual installation.

skbuild/cmaker.py
```diff
@@ -474,17 +474,42 @@ def check_for_bad_installs():
                     ("      " + _install) for _install in bad_installs)
             )))
 
-    def make(self, clargs=(), config="Release", source_dir=".", env=None):
+    def make(self, clargs=(), config="Release", source_dir=".",
+             target="install", env=None):
         """Calls the system-specific make program to compile code.
+
+           target: string
+               Target generated by CMake performing the install.
```
jcfr, line 482:
```
Name of the target responsible to install the project. Default is "install".
.. note::

   To workaround ... [add here comment I suggested below]
```

skbuild/cmaker.py
```diff
@@ -497,12 +522,16 @@ def make(self, clargs=(), config="Release", source_dir=".", env=None):
                 "An error occurred while building with CMake.\n"
                 "  Command:\n"
                 "    {}\n"
+                "  Target:\n"
```
jcfr, line 525: `Target` -> `Install target:`

skbuild/setuptools_wrap.py
```diff
@@ -67,6 +67,11 @@ def create_skbuild_argparser():
         '--cmake-executable', default=None, metavar='',
         help='specify the path to the cmake executable'
     )
+    parser.add_argument(
+        '--target', default=None, metavar='',
```
jcfr, line 71: `target` -> `install-target`

phcerdan:
Split make function into two: `make` and `make_impl` to implement your suggestion:



> Then in case cmake_install_target is specified and different from the default value, should we do this:

        cmd = [self.cmake_executable, "--build", source_dir, "--config", config, "--"]
        cmd.extend(clargs)

> followed by an other call to

        cmd = [self.cmake_executable, "--build", source_dir, "--target", target, "--config", config, "--"]
        cmd.extend(clargs)

> This would ensure the project is always built before building the custom install targets.

phcerdan:
Rebase is needed after your review. I splitted it to facilitate review on the make, make_imp refactor.

phcerdan:
Let me know if I can help to push this forward @jcfr.
I have been deploying 3 projects to pypi using `cmake_install_target` from this fork with no problems. [Example](https://github.com/phcerdan/SGEXT/blob/a41a2e601e674ba41aa3bbe5392b85202c706d1d/deploy/setup.py#L55).

phcerdan:
Hi @jcfr, any hope this would be merged?
Github says Changes Requested, but I would say it's a false positive, I think I managed to deal with all the wise points you raised a year ago.

Let me know how/if I can help further. Thanks!

jcfr:
Thanks for the contributions, the follow up and for the ping. I plan to spend time working on scikit-build this coming week.

Also, we now have some new core contributors helping out with scikit-build projects, and I am beyond grateful having them join the effort.

@henryiii @mayeut 🙏🔥

phcerdan:
@jcfr that's great!

henryiii:
Could you rebase (or merge) with master? Otherwise, I can later.

phcerdan:
@henryiii rebased on top of current master

henryiii:
I think we should be careful about expanding the API surface before #601, but this seems safe enough and all review points addressed. @jcfr could you see if this is good and mark the review complete?

phcerdan:
Thanks @henryiii for taking care of this, please ping me anytime if there is any future issue related with this.

PR #485, opened 2020-04-29
Cygwin Platform Detection

### LINKED ISSUE #484: Cygwin Support?
I am trying to install scikit-build with pip from Cygwin and get the following error message:
```
Unsupported platform: cygwin_nt-10.0-17763. Please contact the scikit-build team.
```

So here I am :)

Platform: GitHub action on `windows-2019` with `Cygwin v3.1.4`.

Proposed PR: #485

ax3l:
Add detection for cygwin.

Close #484

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,7 +104,7 @@ jobs:
     - uses: cygwin/cygwin-install-action@v2
       with:
         platform: x86_64
-        packages: cmake ninja make gcc-g++ gcc-fortran python39 python39-devel python39-pip
+        packages: cmake ninja git make gcc-g++ gcc-fortran python39 python39-devel python39-pip
 
     - name: Install nox
       run: python3.9 -m pip install nox
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,6 +91,27 @@ jobs:
     - name: Test on 🐍 3.10
       run: nox -s tests-3.10
 
+  cygwin:
+    name: Tests on 🐍 3.9 • cygwin
+    runs-on: windows-latest
+    needs: [lint]
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - uses: egor-tensin/setup-cygwin@v3
+      with:
+        platform: x64
+        packages: cmake gcc-g++ python39 python39-pip
+
+    - name: Install nox
+      run: python3.9 -m pip install nox
+
+    - name: Run tests
+      run: python3.9 -m nox -s tests-3.9
+
 
   tests-pypy:
     runs-on: ${{ matrix.runs-on }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,7 +97,7 @@ jobs:
     needs: [lint]
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
       with:
         fetch-depth: 0
 
@@ -113,6 +113,7 @@ jobs:
       env:
         TMP: /tmp
         TEMP: /tmp
+        SETUPTOOLS_SCM_PRETEND_VERSION: "0.14"
       run: python3.9 -m nox -s tests-3.9
 
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,15 +101,18 @@ jobs:
       with:
         fetch-depth: 0
 
-    - uses: egor-tensin/setup-cygwin@v3
+    - uses: cygwin/cygwin-install-action@v2
       with:
-        platform: x64
-        packages: cmake gcc-g++ python39 python39-pip
+        platform: x86_64
+        packages: cmake ninja make gcc-g++ gcc-fortran python39 python39-devel python39-pip
 
     - name: Install nox
       run: python3.9 -m pip install nox
 
     - name: Run tests
+      env:
+        TMP: /tmp
+        TEMP: /tmp
       run: python3.9 -m nox -s tests-3.9
 
 
--- a/skbuild/platform_specifics/cygwin.py
+++ b/skbuild/platform_specifics/cygwin.py
@@ -1,0 +1,33 @@
+"""This module defines object specific to Cygwin platform."""
+
+import sys
+import textwrap
+
+from . import abstract
+from .abstract import CMakeGenerator
+
+
+# pylint:disable=abstract-method
+class CygwinPlatform(abstract.CMakePlatform):
+    """Cygwin implementation of :class:`.abstract.CMakePlatform`."""
+
+    def __init__(self):
+        super(CygwinPlatform, self).__init__()
+        self.default_generators = [CMakeGenerator("Ninja"), CMakeGenerator("Unix Makefiles")]
+
+    @property
+    def generator_installation_help(self):
+        """Return message guiding the user for installing a valid toolchain."""
+        return (
+            textwrap.dedent(
+                """
+            Building Cygwin wheels for Python {pyver} requires Cygwin packages
+            ninja or make and compilers from e.g. gcc-core and gcc-g++.
+            Get them here:
+
+              https://cygwin.com/packages/package_list.html
+            """
+            )
+            .format(pyver="%s.%s" % sys.version_info[:2])
+            .strip()
+        )
--- a/skbuild/platform_specifics/windows.py
+++ b/skbuild/platform_specifics/windows.py
@@ -362,6 +362,7 @@ def _get_msvc_compiler_env(vs_version, vs_toolset=None):
             out = subprocess.check_output(
                 'cmd /u /c "{}" {} {} && set'.format(vcvarsall, arch, vcvars_ver),
                 stderr=subprocess.STDOUT,
+                shell=sys.platform.startswith("cygwin"),
             )
             out = out.decode("utf-16le", errors="replace")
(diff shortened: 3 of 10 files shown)


henryiii:
Any chance you could add a GHA run for this?

henryiii:
(PS: needs rebasing, obviously, to pick up the new CI system)

poikilos:
Due to this issue, [drawing](https://github.com/maoschanz/drawing) is about the 30th program out of 30 attempted that won't bulid on Cygwin over the years☹️. Please merge.

henryiii:
I added a test for Cygwin and it doesn’t work, just this one change is not enough. Happy to have someone suggest the rest of the fix. Not really a windows or Cygwin user.

henryiii:
A few tests could be fixed by resolving the paths (non expanded windows paths, I think). Not sure about the other hard errors though.

.github/workflows/ci.yml
```diff
@@ -91,6 +91,27 @@ jobs:
...
+    - uses: egor-tensin/setup-cygwin@v3
+      with:
+        platform: x64
+        packages: cmake gcc-g++ python39 python39-pip
+
+    - name: Install nox
+      run: python3.9 -m pip install nox
+
+    - name: Run tests
+      run: python3.9 -m nox -s tests-3.9
```
DWesl, line 113:
```suggestion
    - name: Run tests
      env:
        TMP: /tmp
        TEMP: /tmp
      run: python3.9 -m nox -s tests-3.9
```
A lot of the "This path doesn't match that path" problems look like they're in the platform temporary directory.  Changing that to a Cygwin directory might help with that.  Specifying the `shell` might also help.

skbuild/platform_specifics/cygwin.py
```diff
@@ -0,0 +1,33 @@
...
+
+import sys
+import textwrap
+
+from . import abstract
+from .abstract import CMakeGenerator
+
+
+# pylint:disable=abstract-method
+class CygwinPlatform(abstract.CMakePlatform):
```
DWesl, line 11: I inherited from `.unix.UnixPlatform`, but this seems to work just as well.

DWesl (commented):
I found this PR after patching a local copy, and have some suggestions for getting the new CI run to pass.

DWesl:
I think [the last error](https://github.com/scikit-build/scikit-build/runs/6460650356?check_suite_focus=true#step:5:1581) will go away once there's a scikit-build release with this PR in.  That seems [intended by the test](https://github.com/scikit-build/scikit-build/blob/4532855776596b83399535fbe358f7e4a57b57f1/tests/test_setup.py#L328), so maybe the right thing to do is marking that `xfail` until a new release hits PyPI.

The [next error (unknown encoding: mbcs)](https://github.com/scikit-build/scikit-build/runs/6460650356?check_suite_focus=true#step:5:1163) is one I've hit before.  I haven't found a way around it.  Since [that encoding seems baked in to the Windows checks](https://github.com/scikit-build/scikit-build/blob/4532855776596b83399535fbe358f7e4a57b57f1/skbuild/platform_specifics/windows.py#L247-L248), it might be worth skipping [that specific test](https://github.com/scikit-build/scikit-build/blob/4532855776596b83399535fbe358f7e4a57b57f1/tests/test_platform.py#L87-L99) on Cygwin, or marking it an expected failure.  Alternately, the condition `sys.version_info[:3] >= (3, 6, 0)` could add a `sys.platform != "cygwin"` check to avoid the unknown encoding.

The PEP518 test failures I also see locally and don't understand.  The information in the log amounts to "pip install didn't work, because something that isn't pip didn't work", which doesn't narrow down the problem.  These might also go away once a version that supports Cygwin hits PyPI, so maybe this should also be marked `xfail`.  Is there some way to add the currently-tested version of scikit-build to a fake package index for these, or are the tests already doing that?

The [first error](https://github.com/scikit-build/scikit-build/runs/6460650356?check_suite_focus=true#step:5:493) might be related to git?  Adding `git` to the packages to install on Cygwin probably wouldn't hurt.

henryiii:
The PEP 518 test does use a local wheelhouse (see `tests/conftest.py`), but that wheelhouse includes cmake... So same self-dependency problem. I probably could make sure CMake gets installed using the local wheelhouse too, but I think it's fine just to also skip it until we have a Cygwin release.

DWesl:
Should [the attempt to find vcvarsall](https://github.com/scikit-build/scikit-build/blob/4532855776596b83399535fbe358f7e4a57b57f1/skbuild/platform_specifics/windows.py#L359-L362) have `shell=True`?
[The subprocess documentation indicates that is preferred for a string](https://docs.python.org/3/library/subprocess.html#frequently-used-arguments).

henryiii:
Yes it should, that is clearly a shell line.

henryiii:
That's weird, adding shell=True fixed Cygwin and broke Windows. Not sure why, it should have been shell=True in the first place...

skbuild/platform_specifics/windows.py
```diff
@@ -359,6 +362,7 @@ def _get_msvc_compiler_env(vs_version, vs_toolset=None):
             out = subprocess.check_output(
                 'cmd /u /c "{}" {} {} && set'.format(vcvarsall, arch, vcvars_ver),
                 stderr=subprocess.STDOUT,
+                shell=True,
```
henryiii, line 365:
```suggestion
                shell=sys.platform.startswith("Cygwin"),
```

🤦

DWesl:
```pycon
>>> sys.platform
'cygwin'
```
if you decide to go that route.

poikilos: My system (Windows 10) shows sys.platform as "cygwin" in lowercase.

In general, the PR should also document in the readme or somewhere to install python using the Cygwin installer to prevent Cygwin from running the Windows version (otherwise Cygwin uses the Windows version of Python if it is in the PATH (where `sys.platform == "win32"` and `platform.system()=="Windows"` and `platform.platform()==#something like "Windows-10-10.0.19041-SP0"`).

henryiii: It is recommended and [idiomatic](https://docs.python.org/3/library/sys.html#sys.platform) to check `sys.platform` with `startswith`, it used to matter on some systems. I don't think it ever mattered with `cygwin`, but it did on some other platforms (they might be all simpler names since Python 3.8 now, I don't know if they are all covered).

> otherwise Cygwin uses the Windows version of Python

In that case, wouldn't it just build and behave like normal Windows?

henryiii: Subprocess is a little weird on Windows (we have to have special workarounds for bugs in cibuildwheel). I'd be okay to go with the above workaround for now, since it doesn't change the current Windows behavior (though I don't understand why the current behavior works...)

poikilos:
My comment (visible if you click "Outdated") still applies to that commit.

henryiii:
Oops, thanks!

skbuild/platform_specifics/windows.py
```diff
@@ -359,6 +362,7 @@ def _get_msvc_compiler_env(vs_version, vs_toolset=None):
             out = subprocess.check_output(
                 'cmd /u /c "{}" {} {} && set'.format(vcvarsall, arch, vcvars_ver),
                 stderr=subprocess.STDOUT,
+                shell=sys.platform.startswith("Cygwin"),
```
henryiii, line 365:
```suggestion
                shell=sys.platform.startswith("cygwin"),
```

skbuild/platform_specifics/windows.py
```diff
@@ -244,8 +244,11 @@ def _find_visual_studio_2017_or_newer(vs_version):
 
     try:
         extra_args = {}
-        if sys.version_info[:3] >= (3, 6, 0):
-            extra_args = {"encoding": "mbcs", "errors": "strict"}
+        if sys.version_info >= (3, 6):
+            if sys.platform.startswith("cygwin"):
+                extra_args = {"encoding": "utf-8", "errors": "strict"}
+            else:
+                extra_args = {"encoding": "mbcs", "errors": "strict"}
```
henryiii, line 251:
```suggestion
            extra_args = {"encoding": "utf-8" if sys.platform.startswith("cygwin") else "mbcs", "errors": "strict"}
```

skbuild/platform_specifics/windows.py
```diff
@@ -359,6 +362,7 @@ def _get_msvc_compiler_env(vs_version, vs_toolset=None):
             out = subprocess.check_output(
```
henryiii, line 362:
```suggestion
            # TODO: should always be shell=True, but currently requires this
            out = subprocess.check_output(
```

tests/test_setup.py
```diff
@@ -326,6 +327,9 @@ def test_cmake_minimum_required_version_keyword():
     "See https://github.com/conda/conda/issues/508",
 )
 @pytest.mark.skipif(not is_site_reachable("https://pypi.org/simple/cmake/"), reason="pypi.org website not reachable")
+@pytest.mark.xfail(
+    sys.platform.startswith("cygwin"), strict=False, reason="Cygwin needs a release of scikit-build first"
```
DWesl, line 331: Do you want to make this `strict=True` with a note to also update the skipped tests, or just remember to update these once a release gets up to PyPI?

henryiii: I'm okay to just remember - hopefully the release won't be too far in the future, and I'll be trying to remove xfails occasionally.

PR #491, opened 2020-05-07
Fix `add_python_library` and `add_python_extension`

benbovy:
I have the following error when trying to build an extension, even with `find_package(PythonExtensions)` in my `CMakelists.txt`:

```
Unknown CMake command "add_python_extension"
```

I guess this will fix it?


benbovy:
Superseded by #495

PR #493, opened 2020-05-13
Add hello fortran sample package + tests

benbovy:


thewtex:
Cool!

benbovy:
I only tested this PR locally on MacOS (the tests seem to pass on travis too).

I'm not sure what is going wrong on Linux when linking the `_bonjour` module:

```
[ 57%] Generating _bonjourmodule.c, _bonjour-f2pywrappers.f
Reading fortran codes...
	Reading file '/tmp/pytest-of-circleci/pytest-46/test_hello_fortran_wheel0/bonjour/_bonjour.pyf' (format:free)
Post-processing...
	Block: _bonjour
			Block: bonjour
Post-processing (stage 2)...
Building modules...
	Building module "_bonjour"...
		Constructing wrapper function "bonjour"...
		  bonjour()
	Wrote C/API module "_bonjour" to file "./_bonjourmodule.c"
Scanning dependencies of target _bonjour
[ 71%] Building C object bonjour/CMakeFiles/_bonjour.dir/_bonjourmodule.c.o
[ 85%] Building Fortran object bonjour/CMakeFiles/_bonjour.dir/_bonjour-f2pywrappers.f.o
[100%] Linking Fortran shared module _bonjour.cpython-37m-x86_64-linux-gnu.so
bonjour/CMakeFiles/_bonjour.dir/build.make:127: recipe for target 'bonjour/_bonjour.cpython-37m-x86_64-linux-gnu.so' failed
CMakeFiles/Makefile2:190: recipe for target 'bonjour/CMakeFiles/_bonjour.dir/all' failed
```

On windows, it fails as there is no Fortran compiler. Should we skip it? Or should we update the CI setup?

jcfr:
I just rebased the topic so that the Continus integration pipeline re-runs after integrating #494

henryiii:
I've tried adding the fixes from #494, but this still is broken. "fatal error: no input files"

benbovy:
> I've tried adding the fixes from #494, but this still is broken. "fatal error: no input files"

This is likely due to F2PY which doesn't always create Fortran wrappers (i.e., `xxx-f2pywrappers.f`, see the `--wrap-functions` option [enabled by default](https://numpy.org/doc/stable/f2py/usage.html#extension-module-construction)).

In the test program used here, there's no Fortran function so none of those wrapper source files are created.

I'm not sure what would be a good fix, since CMake doesn't know in advance whether or not those wrapper file will be created when running the F2PY command. Is there some way in CMake to check that or just skip non-generated source files?

benbovy:
Also, CI failure on Windows seems related to this: https://gitlab.kitware.com/cmake/cmake/-/issues/17810

henryiii:
Is it realistic to have no Fortran function? Maybe the first step would be to have a Fortan function, then a second step could solve the functionless case?

henryiii:
We can disable the test on Windows for now. We'll need to solve this, but might need some support from upstream.

benbovy:
> Is it realistic to have no Fortran function? 

Yes, at least in my case, but I don't know if it's representative of the general cases.

> Maybe the first step would be to have a Fortran function

I've tried that locally with a dummy function, but there are still some wrappers that are not generated. More specifically, `"${CMAKE_CURRENT_BINARY_DIR}/${_name}-f2pywrappers2.f90"` is still missing. I'm not sure in which cases this file is generated.

Maybe we can skip using explicitly `add_f2py_target`'s `OUTPUT_VAR` option in the `bonjour` test module here and use instead the module name like in the [example shown in the documentation](https://scikit-build.readthedocs.io/en/latest/cmake-modules/F2PY.html#id1)? It works but it gives a CMake warning (no explicit source files given to `add_library`).

We could also clarify in the docs that `OUTPUT_VAR` contain the path to the C module source file *and* some Fortran wrapper source files where the latter may or may not exist after running F2PY, depending on the content of the signature file... if that's something that users should care about rather than something to fix in scikit-build.

PR #495, opened 2020-05-13
Misc. fixes for F2PY and PythonExtensions modules

### LINKED ISSUE #492: Troubles with F2PY and PythonExtensions cmake modules
I'm having trouble using those modules to build and install a Python module from Fortran code (I'm using version 0.11.1).

I realize that #491 is included in some fix commits found here:

https://github.com/scikit-build/scikit-build/compare/master...jcfr:add-UseF2PY-and-UsePythonExtensions

I think those commits would also fix the other issues that I currently have. For some reason, those fixes are not present in the master branch.

benbovy:
Closes #492 

TODO:

- [ ] add some tests.

### DIFF
--- a/skbuild/resources/cmake/FindF2PY.cmake
+++ b/skbuild/resources/cmake/FindF2PY.cmake
@@ -93,7 +93,9 @@ add_library(_f2py_runtime_library STATIC ${_f2py_sources})
 target_include_directories(
   _f2py_runtime_library
   PRIVATE ${PYTHON_INCLUDE_DIRS} ${NumPy_INCLUDE_DIRS}
-)
+  )
+
+set_target_properties(_f2py_runtime_library PROPERTIES POSITION_INDEPENDENT_CODE ON)
 
 set(F2PY_LIBRARIES _f2py_runtime_library)
 set(F2PY_INCLUDE_DIRS "${F2PY_INCLUDE_DIR}" "${NumPy_INCLUDE_DIRS}")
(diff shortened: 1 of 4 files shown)


jcfr:
I just rebased the topic so that it re-run after integrating #494

jcfr:
Thanks for the fixes :pray: 

These are all sensible changes :+1: 

Do not hesitate to split the set of changes into few independent commits.

I think adding a sample project `hello-fortran` would make sense for the testing. I have no issue with tracking and addressing this independently of this PR.

benbovy:
> Do not hesitate to split the set of changes into few independent commits

Do you want me to do this in a new PR?

> I think adding a sample project hello-fortran would make sense for the testing

#493 does not currently include an example that uses the functions in `UsePythonExtensions`. It would be nice to test these too. Should we wait #493 to be merged and then add another subdirectory in `hello_fortran`?

anlach-nnl:
These changes would be super useful to me.  They seem like good fixes made by @benbovy.  Any update on the status of this, @jcfr ?  What still needs to be done?

jcfr:
@anlach-nnl  Thanks for reaching out, in the middle of "travelling" and tackling some other deadlines, I have a limited bandwidth. I will be able to have a look only after October 11th. Would that be okay ?

anlach-nnl:
@jcfr Yes, that works for me.  I have it working with a local build, and I was just wondering when I would be able to get it from conda-forge.  I'm glad it's on your radar!  Thanks for the quick response

henryiii:
From what I understand, this PR fixes broken tests in #493? (Broken meaning they expose broken behavior) Shouldn't those be added to this PR?

PR #498, opened 2020-05-20
setuptools_wrap: Remove cmake_process_manifest_hook keyword from argument dict

Jmennius:
So that setuptools won't complain about unknown setup options.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,12 @@ details, see the commit logs at http://github.com/scikit-build/scikit-build
 Next Release
 ============
 
+Bug fixes
+---------
+
+* Fixed a regression that caused setuptools to complain about unknown setup option
+  (`cmake_process_manifest_hook`).
+
 Documentation
 -------------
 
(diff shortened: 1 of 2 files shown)


Jmennius:
Need #494 for tests to pass (I'll rebase then).

jcfr:
> Need #494 for tests to pass (I'll rebase then).

Thanks for your contribution  :pray: 

Few nitpicks:
* [ ] add entry to [CHANGES](https://github.com/scikit-build/scikit-build/blob/master/CHANGES.rst)
* [ ] update commit message. See below for a suggestion
* [ ] bonus: any test that could be improved to prevent future regression like this one ? 

```
setuptools_wrap: Fix unknown setup options related to cmake_process_manifest_hook keyword

This commit addresses a regression introduced in 66d7a32  (Add hook to process
cmake install manifest) and prevents setuptools from complaining about unknown
setup options.

It removes cmake_process_manifest_hook keyword from argument dict.
```

Jmennius:
Thanks for the feedback!

I've implemented first two points but I'll take a look at tests later :)

jcfr:
Will integrate after Continuous integration pipeline completes.

PR #508, opened 2020-08-26
Proposal: Link to latest cmake documentation

Eothred:
Propose to link to latest version of cmake documentation rather than a specific version


jcfr:
Thanks for the contribution :pray: 

Could you also add an entry in the changelog, see https://github.com/scikit-build/scikit-build/blob/master/CHANGES.rst#documentation

PR #512, opened 2020-10-06
add openbsd platform

kdc-rgc:
added a check for the "openbsd" platform


thewtex:
@kdc-rgc this looks good. Did you test it on an OpenBSD system? Why was the PR closed?

CC: @jcfr

PR #515, opened 2020-10-23
fix: hide warning if SKBUILD is not used

henryiii:
I believe this should hide the warning whenever SKBUILD is not checked in the CMake files. It also isn't a user-facing variable (that is, users don't set it).


jcfr:
Very neat. Thanks for address this. Would you mind adding an entry to the CHANGE file ?

henryiii:
Sure, done. This one does have a CHANGES file, unlike `cmake_python_distributions`. :)

jcfr:
Thanks.

Re: cmake distribution changes
Good point I got confused.

PR #516, opened 2020-10-23
feat: support cmake_findpython

henryiii:
Starting basic support for FindPython (not in the extra module just yet, just in skbuild itself). Happy to add tests, where should they go?


henryiii:
Would there be any harm in just passing both sets of variables, and setting them to INTERNAL?

henryiii:
Also, the library may not be present (such as in manylinux), so this should be handled nicely.

PR #526, opened 2021-01-03
Support Visual Studio 2019

### LINKED ISSUE #496: Support for Visual Studio 2019?
Is visual studio 2019 supported and is there a plan to support?

Using the latest scikit-build 0.11.1 and `pip install` works on a system with Visual Studio 2017 and fails on 2019 with long error message about missing 2017. Both tested on Windows 10 Home.

Specifically:
```
python -m pip install Tasmanian --user
```

I have also tried explicitly adding the generator, but it did not help:
```
python -m pip install Tasmanian --user --global-option="-G \"Visual Studio 16 2019\""
python -m pip install Tasmanian --user --global-option='-G "Visual Studio 16 2019"'
python -m pip install Tasmanian --user --global-option='-G Visual Studio 16 2019'
```
Note that pip works fine with VS2017 and the project compilers fine with VS2019 outside of the pip environment.

YannickJadoul:
On GitHub Actions' `windows-latest` image VS 2019 is available, but not detected by scikit-build. Python >= 3.5 can be built by VS 2019, though (cfr. https://wiki.python.org/moin/WindowsCompilers).

Tested on in pybind11's `scikit_build_example` example repository:

- https://github.com/YannickJadoul/scikit_build_example/actions/runs/458497089
- https://github.com/YannickJadoul/scikit_build_example/actions/runs/458500361

Note in particular the removal `ilammy/msvc-dev-cmd@v1` action in https://github.com/YannickJadoul/scikit_build_example/commit/2c3480a01e971b1d6b460f6287d0215fee088ca1, which masked/worked around the failure to detect VS 2019.

Fixes #496.

(CC @henryiii)


henryiii (dismissed):
Looks very reasonable to me!

YannickJadoul:
Fixed the AppVeyor build (you might want to steal this for `master`, since it was already broken there; feel free to do so).

Azure still needs some investigation/debugging, I believe.

jcfr:
Thanks for the contribution 🙏 Will review and integrate later today.

YannickJadoul:
Great! I'm still trying to figure out why some tests are failing, but any hints are of course welcome :-)

YannickJadoul:
(Apologies for debugging here; I can't reproduce on my test project on GHA)

jcfr:
Update:
* Working on integrating fixes from https://github.com/scikit-build/scikit-build/pull/526/commits/6e8b96b929f511862c123d426759da877532cc76 (Add -T and -A flags in cmaker.py). Just added a test (expected to fail) checking if toolset CMake argument `-T` is passed down to the CMake project: https://github.com/scikit-build/scikit-build/pull/536

thewtex:
While this issue should be addressed, it is worth noting for folks looking for a solution that there is a workaround to call the Visual Studio environment setup script beforehand. Example:

https://github.com/InsightSoftwareConsortium/ITKCuberille/blob/9ba9620d9abb25dc511d8d437fb7d1c7bda3025d/.github/workflows/build-test-package.yml#L67

A side note: Visual Studio ships with `ninja.exe` now. We may want to make that the default CMake Generator on Windows -- it is so much faster and more scalable than Visual Studio project solution files.

jcfr:
> Visual Studio ships with ninja.exe now

Do you happen to know which version of VS started to ship with  it ? 

> We may want to make that the default CMake Generator on Windows

The good news is that scikit-build is always trying to lookup ninja first. See https://github.com/scikit-build/scikit-build/blob/4e2928d93ba275f5cfc3837c174c25e6c4a73ac0/skbuild/platform_specifics/windows.py#L91

Idea: It would be helpful to improve CMake itself to have a generator like `Visual Studio 16 2019 Ninja` (or similar)

jcfr:
Update 2:  I started to actively working on integrating VS2019 support and plan to make more progress next week

henryiii:
I've rebased with the new CI. Two issues pop out (other than general unrelated warnings we should address later): On the 2017 image, it's trying to find 2019 and failing. And on the 2019 image, it's failing too.

jcfr:
Thanks for taking care of rebasing. On my list is to finalize the vs2019 support, this is now in the critical path of few projects.

thewtex:
> Do you happen to know which version of VS started to ship with it ?

VS 2019. Folks that Meson uses ninja by default -- it is so much faster. It is worth considering ninja a default for skbuild / VS 2019.

henryiii:
In  the meantime, could you update the documentation to list VS 2019 ? (see this for example of text ... https://github.com/scikit-build/scikit-build/compare/master...jcfr:496-support-vs2019

_Originally posted by @jcfr in https://github.com/scikit-build/scikit-build/issues/557#issuecomment-879121871_

jcfr:
> It is worth considering ninja a default for skbuild / VS 2019.

Agreed. It turns out that Ninja is already the generator looked up first and the preferred way.

henryiii:
It will be a little bit before I can get back to this; currently, one of the problems is that the tests are expecting the VS generator and are getting Ninja instead, I believe.

skbuild/platform_specifics/windows.py
```diff
@@ -58,31 +58,37 @@ def __init__(self):
                 "https://www.microsoft.com/download/details.aspx?id=8279"
             )
 
-        # For Python 3.5: VS2017, VS2015
+        # For Python 3.5: VS2019, VS2017, VS2015
         elif version.major == 3 and version.minor == 5:
-            supported_vs_years = [("2017", "v140"), ("2015", None)]
+            supported_vs_years = [("2019", "v142"), ("2017", "v141"), ("2015", None)]
```
henryiii, line 63: Not sure why this was "v140", but might be needed that way...

henryiii:
```suggestion
            supported_vs_years = [("2019", "v142"), ("2017", "v140"), ("2015", None)]
```

YannickJadoul: As far as I figured out, `v140` is needed to build against the toolset of the MSVC version in VS2015.

https://wiki.python.org/moin/WindowsCompilers is not entirely clear about this, but if you look at [Python 3.6.8 for Windows](https://www.python.org/ftp/python/3.6.8/python-3.6.8-embed-amd64.zip), you see it only contains `vcruntime140.dll`. Compare to e.g. [3.9.6](https://www.python.org/ftp/python/3.9.6/python-3.9.6-embed-amd64.zip), which also contains `vcruntime140_1.dll` (which I'm guessing corresponds to `v141`, but as we know, VS/MSVC's version numbering and ABI compatibility are not all too straightforward to understand).

YannickJadoul: So, `("2019", "v142")`: maybe that should also have `v140`? But then it also seems to depend on which version of Python you're building for, if I understood correctly?

henryiii: That's mentioned here: https://cibuildwheel.readthedocs.io/en/stable/faq/#importerror-dll-load-failed-the-specific-module-could-not-be-found-error-on-windows

If you use 2019, there's an extra include, and but you can disable it with a flag.

henryiii: > VS/MSVC's version numbering and ABI compatibility are not all too straightforward to understand 

+1!

YannickJadoul: Ha, you're right. So `vcruntime140_1.dll` does _not_ correspond to `v141`, indeed. Any idea why you had to make the `v140` -> `v141` change, then?

One thing to definitely test is the VS2019 v142 in combination with the oldest Python version available, perhaps? If v142 works on that, then my guess about setting `v140` for backwards compatibility was definitely wrong.

YannickJadoul: Ah, wait. Another online search, and I found this: https://docs.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017?view=msvc-160

> You can mix binaries built by different versions of the v140, v141, and v142 toolsets. However, you must link by using a toolset at least as recent as the most recent binary in your app.

henryiii:
Current status, since my son has commandeered the Windows box I was working on: On MSVC 2019, I get a loop like this:

```
-- Configuring done
-- Generating done
-- Build files have been written to: C:/Users/runneradmin/AppData/Local/Temp/pytest-of-runneradmin/pytest-248/setup_requires_keyword_include0/_skbuild/win-amd64-3.5/cmake-build
[0/1] Re-running CMake...

-- Configuring done
-- Generating done
-- Build files have been written to: C:/Users/runneradmin/AppData/Local/Temp/pytest-of-runneradmin/pytest-248/setup_requires_keyword_include0/_skbuild/win-amd64-3.5/cmake-build
[0/1] Re-running CMake...
```

(I think only in CI, not getting that locally)

On MSVC 2017, it seems to be trying 2019 then dying, not moving on to 2017. It's possible this logic here has been broken for a while, as we mostly probably use MSVC 2017.

henryiii:
It's succeeding on "`Trying "NMake Makefiles JOM (Visual Studio 16 2019 x64 v142)" generator - success`" but then breaking because it has no idea what `cl` is.

henryiii:
I'm removing the "NMake Makefiles JOM" from the default generators list - the documentation clearly says it is not in the default generators, and it's misbehaving. The other three generators are behaving correctly (well, except for Ninja on MSVC 2019 breaking under one test, which might be related to the date of the pytest generated file or something similar)

henryiii:
Okay, this is ready to go (I'm pulling out a couple of unrelated changes and will make PRs for those). There's a bug in our test suite where you can't run it twice with two different versions of Python when it's using the Ninja generated on Windows- the second run breaks Ninja. It's only on a single, "setup_requires" test that generates a CMakeLists. I've just split the job for now.

After these go in I'd be ready for a 0.12. :)

henryiii:
Going to merge, then I want to see if we can enable testing on 3.10. If we can pass on the release of 3.10rc1 later today, then we can add 3.10 to our supported version classifiers :).

PR #527, opened 2021-01-04
Add support for Visual Studio 2019

jorgen:


henryiii:
Already being worked on in #526.

Issue #529, opened 2021-02-06
ARCHFLAGS not picked up to build Apple Silicon arm64 and universal2 wheels


YannickJadoul:
`cibuildwheel` has rolled out support for Apple Silicon wheels (both `arm64` as well as the combined `universal2`), but scikit-build passes ` -DCMAKE_OSX_ARCHITECTURES=x86_64` to CMake, and thus (silently!) fails to produce wheels with a different architecture.

See also this thread: https://github.com/joerick/cibuildwheel/discussions/579#discussioncomment-345617
CC @henryiii

thewtex:
CC @jcfr

henryiii:
The cmake package needs to be updated to pick up the universal2 official CMake release, as well. I was waiting till Pip supported Universal2 wheels (21.0.1), but now will likely be out for two weeks starting very soon due to the upcoming birth of my daughter.

henryiii:
Cross-ref: https://github.com/scikit-build/cmake-python-distributions/pull/123

YannickJadoul:
Hurray! :balloon: Thanks for taking further care of this, @henryiii :-)

PR #530, opened 2021-02-14
Respect CMAKE_OSX_ARCHITECTURES and correctly produce universal2 macOS wheels

YannickJadoul:
A first, partial fix for #529. This PR does two things: 1) it makes sure that `CMAKE_OSX_ARCHITECTURES` is not overridden if set externally, and 2) it makes sure that the correct wheel file name is given for `universal2` wheels (when `CMAKE_OSX_ARCHITECTURES` contains both `arm64` and `x86_64`).

### DIFF
--- a/skbuild/constants.py
+++ b/skbuild/constants.py
@@ -32,8 +32,18 @@ def _default_skbuild_plat_name():
         release = os.environ.get("MACOSX_DEPLOYMENT_TARGET", release)
         split_ver = release.split('.')
         machine = os.environ.get("CMAKE_OSX_ARCHITECTURES", machine)
+
+        # Handle universal2 wheels, if two architectures are requested.
         if set(machine.split(';')) == {'x86_64', 'arm64'}:
             machine = 'universal2'
+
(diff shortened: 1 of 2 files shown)


YannickJadoul:
CC @henryiii.

maxmouchet:
Hi,

Any update on merging this?
Building with `CIBW_ARCHS_MACOS: universal2` still produces an `x86_64` wheel.

Thanks for this project!

henryiii:
Package updates for AS need to go in the order cmake -> ninja -> scikit-build, currently the hold-up is ninja (cmake 3.20.2 with full AS support was just released a few days ago). We'll try to work on these PRs once that goes through.

henryiii (approved):

henryiii:
This is only part way there, but a bit of a "minimal needed" step to get there. I think we should also support the "standard" variables (like ARCHFLAGS) so that a Scikit-Build projects behaves just like a setuptools project - but that can be done later.

I'm going to update the changelog in a batch so there aren't conflicts.

PR #538, opened 2021-05-13
Handle updates to CYTHON_FLAGS from CMake code

oiffrig:
At the moment, CYTHON_FLAGS is read only once, when `include(UseCython)` is called. This means further changes to this variable from CMake code are not taken into account. This PR allows users to update CYTHON_FLAGS as needed (for instance when defining multiple modules with different flags).


henryiii (approved):
This looks like an excellent, well tested PR to me.

PR #540, opened 2021-06-08
feat: nox

henryiii:
Adding support for nox, along with a bit of cleanup. @jcfr, I'd like to remove `Makefile`, `tox.ini`, and maybe `scikit-ci.yml` in favor of nox as well.

Nox is much simpler than tox, providing a clear, easy to read list of instructions rather than the somewhat magical incantations of tox that do not help a user to understand what is going on. Nox prints the exact commands in nice colors as it goes, in fact. Unlike make, it is integrated with virtual environments, so there's no setup required and no cleanup needed afterwards. The downside is it's a bit slower (a second or two is spent on preparing a virtual env), but repeatability is paramount for new users, not speed. Daily developers don't need to use nox on the tasks they are doing often (though I leave docs to nox pretty much always).

We don't have to use it in the CI, but we can.

I also want to move the linting driver to pre-commit in the near future, but we can allow that to be run from nox as a helper for new contributors, too.


jcfr:
Thanks for introducing nox and for taking the time to summarize it.

> [...]  than the somewhat magical incantations of tox that do not help a user to understand what is going on.

While I liked the declarative nature of tox, in practice it has been a bit challenging to work with.  

> Nox prints the exact commands in nice colors as it goes, in fact. Unlike make, it is integrated with virtual environments, so there's no setup required and no cleanup needed afterwards. 

This is huge :100: 

> The downside is it's a bit slower (a second or two is spent on preparing a virtual env), but repeatability is paramount for new users, not speed. 

Agreed, not a problem at all .. .especially considering the test suite take longer in comparison.

Out of curiosity, is the new environment setup once per test run ? Or for each tests with a run ? 

> Daily developers don't need to use nox on the tasks they are doing often (though I leave docs to nox pretty much always).

:+1: 

For future reference: https://nox.thea.codes/en/stable/index.html

henryiii:
It's one environment per session. So there's one environment made for Python X.Y that then runs all the tests for Python X.Y. But the lint session uses a new environment, even if it happens to be the same. Just like pytest, you can parametrize any session (say, over architectures), then it's one for each parametrization. You can only run one item from a parametrization (for example, `nox -s tests-3.9` will only run the Python 3.9 tests, regardless of what. versions of Python you have locally)

There is a way to request caching for an environment, but I've mostly left it off for robustness. I _think_ it's probably okay to enable caching for any session that doesn't include a `session.install(".")` in it, like linting and various scripts. But haven't played with it to see what's best there.

My Scikit-HEP developer page on Nox finally passed review and is up: https://scikit-hep.org/developer/tasks

PR #544, opened 2021-06-19
fix: fix regex in test_get_python_version for Python 3.10

mgorny:
Fix the regular expression in test_get_python_version to permit
the minor version to contain more than one digit.  This fixes the test
on Python 3.10.

### DIFF
--- a/tests/test_cmaker.py
+++ b/tests/test_cmaker.py
@@ -21,7 +21,7 @@ from . import _tmpdir, get_cmakecache_variables
 
 
 def test_get_python_version():
-    assert re.match(r'^[23](\.?)[0-9]$', CMaker.get_python_version())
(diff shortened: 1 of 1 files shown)


henryiii:
Fantastic, thanks! Are you testing on 3.10? Just wondering if there are any other places where this sort of mistake might have slipped in. We'll likely turn on testing for 3.10 once we can produce wheels for cmake/ninja, which is around the time of the final beta / first RC IIRC (needs to be ABI stable WRT 3.10 final).

henryiii (approved):

mgorny:
Yeah, Gentoo has optional support for Python 3.10, so I'm moving through packages and testing them. Ideally, I'd like everything to be ready before 3.10 final. AFAIK Fedora Rawhide is also testing everything on Python 3.10.

PR #549, opened 2021-07-02
ci: move most CI to GHA

### LINKED ISSUE #547: ci: CI broken due to Scikit-CI
Scikit-CI is broken on Python 2. Let's continue with the cleanup suggested in #540.

henryiii:
Closes #547. Removes Travis, CircleCI, and Appveyor. Moves Azure to using nox instead of scikit-ci. Adds GHA test run. For the moment, coverage is not included, can be added later. Drops Makefile and tox.ini.


jcfr (approved):

jcfr:
After addressing the failures, transitioning part or all of the build to GHA is very sensible. Thanks for helping with this :pray:

henryiii:
It's interesting that the ubuntu tests take about a minute to run, while macOS is more like 5 minutes (and Windows is even longer, I think, but that's slightly more to be expected).

PR #552, opened 2021-07-02
ci: include search vars in GHA too

henryiii:
At this point, I think it's redundant with Azure. But we can keep it, no harm.


jcfr:
>  I think it's redundant with Azure. But we can keep it, no harm.

Ditto. Increasing the platform we test on is good. Since in practice, scikit-build will be used on GHA, Azure, etc ... this is great.

PR #555, opened 2021-07-02
feat: better Apple Silicon support

henryiii:
This improves the support for Apple Silicon by respecting the setting of ARCHFLAGS, just like setuptools does. Also fixed the output of names like "11.4" for the macOS version; for versions 11+, only the major version is used, and the minor version should remain 0. (Apple is now incrementing the major version every release, instead of the minor version, which is no longer stuck at 10).

Followup to #530.


henryiii:
Since this has been a week, let's merge, then I'll try it out in a draft PR to pybind/scikit_build_example.

PR #556, opened 2021-07-09
fix: archflags was not being parsed correctly

henryiii:
Followup to #555. The format is `-arch <arch>`, which was in some version of the PR I think but didn't make into the final version. Using a simpler method than my original one, anyway. Testing in https://github.com/henryiii/scikit_build_example/pull/3 (which was actually supposed to target pybind/scikit_build_example, but this is fine for a test.

### DIFF
--- a/skbuild/constants.py
+++ b/skbuild/constants.py
@@ -33,7 +33,7 @@ def _default_skbuild_plat_name():
     # it returns the macOS version on which Python was built which may be
     # significantly older than the user's current machine.
     release = os.environ.get("MACOSX_DEPLOYMENT_TARGET", release)
-    major_macos, minor_macos = release.split('.')[:2]
+    major_macos, minor_macos = release.split(".")[:2]
 
     # On macOS 11+, only the major version matters.
     if int(major_macos) >= 11:
@@ -42,14 +42,17 @@ def _default_skbuild_plat_name():
     # Use CMAKE_OSX_ARCHITECTURES if that is set, otherwise use ARCHFLAGS,
     # which is the variable used by Setuptools. Fall back to the machine arch
     # if neither of those is given.
-    machine = os.environ.get("ARCHFLAGS", machine)
+    archflags = os.environ.get("ARCHFLAGS")
+    if archflags is not None:
+        machine = ";".join(set(archflags.split()) & {"x86_64", "arm64"})
+
     machine = os.environ.get("CMAKE_OSX_ARCHITECTURES", machine)
 
     # Handle universal2 wheels, if those two architectures are requested.
-    if set(machine.split(';')) == {'x86_64', 'arm64'}:
-        machine = 'universal2'
+    if set(machine.split(";")) == {"x86_64", "arm64"}:
+        machine = "universal2"
 
-    return 'macosx-{}.{}-{}'.format(major_macos, minor_macos, machine)
+    return "macosx-{}.{}-{}".format(major_macos, minor_macos, machine)
 
 
 _SKBUILD_PLAT_NAME = _default_skbuild_plat_name()


skbuild/constants.py
```diff
@@ -42,14 +42,17 @@ def _default_skbuild_plat_name():
     # Use CMAKE_OSX_ARCHITECTURES if that is set, otherwise use ARCHFLAGS,
     # which is the variable used by Setuptools. Fall back to the machine arch
     # if neither of those is given.
-    machine = os.environ.get("ARCHFLAGS", machine)
+    archflags = os.environ.get("ARCHFLAGS")
+    if archflags is not None:
+        machine = ";".join(set(archflags.split()) & {"x86_64", "arm64"})
```
jcfr, line 47: may be sensible to introduce an intermediate variable `supported_macos_architectures={"x86_64", "arm64"}`  ?

henryiii: Sure, probably `SUPPORTED_MACOS_ARCHETECTURES` if it's a constant. I was tempted to do this originally, but it would not have simplified the universal line below, because that is uniquely these two archs, not all supported archs (which currently is just the same two, of course).

jcfr: Since these are only used in this function ... may be this would make things more explicit ?


```diff
    # Use CMAKE_OSX_ARCHITECTURES if that is set, otherwise use ARCHFLAGS,
    # which is the variable used by Setuptools. Fall back to the machine arch
    # if neither of those is given.
    archflags = os.environ.get("ARCHFLAGS")
    if archflags is not None:
+       supported_macos_architectures = {"x86_64", "arm64"}
-        machine = ";".join(set(archflags.split()) & {"x86_64", "arm64"})
+        machine = ";".join(set(archflags.split()) & supported_macos_architectures)

    machine = os.environ.get("CMAKE_OSX_ARCHITECTURES", machine)

    # Handle universal2 wheels, if those two architectures are requested.
+    macos_universal2_architectures = macos_architectures
-    if set(machine.split(";")) == {"x86_64", "arm64"}:
+    if set(machine.split(";")) == macos_universal_architectures:
        machine = "universal2"
```

May also be sensible to update the docstring of `_default_skbuild_plat_name`

henryiii: That looks good. Not sure I'd do `macos_universal2_architectures = macos_architectures` over `macos_universal2_architectures = {"x86_64", "arm64"}` - if you were to add something to `macos_architectures`, it would not be added to `macos_universal2_architectures`. Keeping code DRY is good, but not if the abstraction is wrong - [AHA](https://testandcode.com/160).

jcfr: Thanks for the AHA reference, I will read it through :pray: 

What about ?

```diff
    # Use CMAKE_OSX_ARCHITECTURES if that is set, otherwise use ARCHFLAGS,
    # which is the variable used by Setuptools. Fall back to the machine arch
    # if neither of those is given.
+   supported_macos_architectures = {"x86_64", "arm64"}
    archflags = os.environ.get("ARCHFLAGS")
    if archflags is not None:
-        machine = ";".join(set(archflags.split()) & {"x86_64", "arm64"})
+        machine = ";".join(set(archflags.split()) & supported_macos_architectures)

    machine = os.environ.get("CMAKE_OSX_ARCHITECTURES", machine)

    # Handle universal2 wheels, if those two architectures are requested.
-    if set(machine.split(";")) == {"x86_64", "arm64"}:
+    if set(machine.split(";")) == supported_macos_architectures:
        machine = "universal2"
```

henryiii: I would do:

```diff
    # Use CMAKE_OSX_ARCHITECTURES if that is set, otherwise use ARCHFLAGS,
    # which is the variable used by Setuptools. Fall back to the machine arch
    # if neither of those is given.
+   supported_macos_architectures = {"x86_64", "arm64"}
    archflags = os.environ.get("ARCHFLAGS")
    if archflags is not None:
-        machine = ";".join(set(archflags.split()) & {"x86_64", "arm64"})
+        machine = ";".join(set(archflags.split()) & supported_macos_architectures)

    machine = os.environ.get("CMAKE_OSX_ARCHITECTURES", machine)

    # Handle universal2 wheels, if those two architectures are requested.
+   macos_universal2_architectures = {"x86_64", "arm64"}
-   if set(machine.split(";")) == macos_universal2_architectures:
        machine = "universal2"
```

Or,  maybe, you could set `supported_macos_architectures = macos_universal2_architectures` - this direction is correct. If Apple switches to a new architecture some day, it is clear you need to now add one to `macos_universal2_architectures` to get `supported_macos_architectures`.

PR #563, opened 2021-07-31
chore: run pre-commit with nox

henryiii:


henryiii:
This is just a fix for something I either forgot to change or forgot we already supported.

PR #568, opened 2021-08-03
docs: update changelog

henryiii:


henryiii:
We are now passing all tests in https://github.com/pybind/scikit_build_example/pull/16 - including new windows ones and windows-latest. I think we should merge and release!

PR #569, opened 2021-08-03
chore: bump to version 0.12.0

henryiii:


henryiii:
This is working in both https://github.com/scikit-build/scikit-build-sample-projects/pull/15 and https://github.com/pybind/scikit_build_example/pull/16, allowing old workarounds to be (optionally) dropped.

Issue #573, opened 2021-08-04
Update requirements files to fix info reported by requires.io


jcfr:
As of August 4th, visiting  https://requires.io/github/scikit-build/scikit-build/requirements/?branch=master reports that setuptools and sphinx are outdated. For future reference, see screenshot below.

* `setuptools`: Add `  # rq.filter: <45` (see documentation of requires filter [here](https://requires.io/features/)). Indeed, as explained in  https://github.com/scikit-build/scikit-build/commit/a0cbcedea94514a92f3f9230825379fca2128fbd, an older setuptools needs to be used for python < 3.

* `sphinx`: @henryiii , in https://github.com/scikit-build/scikit-build/commit/9087b5defc1d4c29ec6933fea4fd20d26468dc04 you added a constraint for not using the latest sphinx. Do you recall the rational for doing so ?

---------------

#### scikit-build readme file

![image](https://user-images.githubusercontent.com/219043/128127925-f40d8991-38fb-496a-9486-1a7a971efa79.png)

#### requires.io website

![image](https://user-images.githubusercontent.com/219043/128127840-f758355a-1cb6-48fd-8ad7-c432e40e1261.png)

henryiii:
### Setuptools

This is a false positive; we pin setuptools only for Python 2, since 45+ is Python 3 only. Since Pip 8, this has been handled correctly by Pip; it will read the Requires-Python metadata slot and select the most recent compatible setuptools, so really, we could drop it and be just fine today. But if we plan to drop Python 2 support in 4 months or less, then not sure it's worth changing. I'd be fine either way - I don't pin like this in any other package, and have had no problems.

### Sphinx

This is docs-only, so it does not affect users. And Sphinx is very slow to be supported by all it's plugins - IIRC, sphinx-rtd-theme is behind (it usually is) for the update. It might have only been warnings, though, don't remember, from looking at the theme it seems it just produces a bunch of warnings with Sphinx 4.

[...]

PR #574, opened 2021-08-05
docs: use moderncmakedomain, Sphinx 4

henryiii:
This enables the use of Sphinx 4. The copy of `docs/cmake.py` from CMake has been replaced with `sphinxcontrib-moderncmakedomain`, which is a port of that file to PyPI, and is much more up-to-date (from CMake 3.19, whereas our copy was much other and didn't support Sphinx 4). Really, it seems like something like that could be maintained as part of scikit-build; curious to know @slurps-mad-rips's opinion on that. We use `sphinxcontrib-moderncmakedomain` in pybind11.

This also pins docutils, as docutils 0.17 is not compatible with sphinx-rtd-theme currently. I'd really like to move away from the rtd theme for this and other reasons.

Requested in #573.


henryiii:
@jcfr or @thewtex , could you go to ReadTheDocs and activate pull request previews? Admin -> Advanced Settings -> Build pull requests for this project. Then there will be a link to preview the docs in the PR.

jcfr:
>  Admin -> Advanced Settings -> Build pull requests for this project.

All set. To streamline maintenance, could I add you as maintainer on readthedocs ? what is your username ?

henryiii:
I'm `henryiii` almost everywhere (twitter being the only exception I can think of ATM).

jcfr:
Since the push event didn't make it ... I just re-synced the webhook. I am also wondering if the preview would be available only for new PR ?

image

henryiii:
There it does, it's in the list of checks now. It can be a little finicky - resyncing the webhook after enabling it was a good idea.

slurps-mad-rips:
> Really, it seems like something like that could be maintained as part of scikit-build

I'm totally fine with this, and would be willing to transfer ownership/maintainership of the repository and the package on PyPI.

I had originally set it up to help write CMake documentation, but the cmake domain that kitware uses is a bit wonky and expects certain declarations to be in specific locations (at least the last time I tried to use it) and just never ended up using it as a result 😅

henryiii:
Interesting, this had a syntax highlighting problem here: https://scikit-build--574.org.readthedocs.build/en/574/cmake-modules/PythonExtensions.html#example-usage - I noticed the warning when building, but it does look like it's unable to render this CMake snippet, while the current file can.

jcfr:
> noticed the warning when building, but it does look like it's unable to render this CMake snippet

For reference,  here is the warning:

```
/home/jcfr/Projects/scikit-build/skbuild/resources/cmake/FindPythonExtensions.cmake:169: WARNING: Could not lex literal_block as "cmake". Highlighting skipped.
```

jcfr:
> docs: Fix CMake example syntax in FindPythonExtensions documentation

Worth noting that pygment is not aiming to report syntax error and expect the block to be valid.

Quote from https://pygments.org/faq/#can-i-use-pygments-for-programming-language-processing

> Also, error reporting is not the scope of Pygments. It focuses on correctly highlighting syntactically valid documents, not finding and compensating errors.

jcfr:
> I'm totally fine with this, and would be willing to transfer ownership/maintainership of the repository and the package on PyPI

@slurps-mad-rips  Thanks for considering this :pray:  

I am currently discussing with the CMake team and will follow up shortly with an update.

@henryiii  Do  you think  including the update of the `moderncmakedomain` as part of the `cmake-python-distributions` release process something reasonable ? 

In a nutshell, this would consist in copying the `cmake.py` from CMake code base, while would trigger the continuous integration and automatic update to PyPI (based on GitHub Action) that we would be adding to the project.

skbuild/resources/cmake/FindPythonExtensions.cmake
```diff
@@ -201,7 +201,7 @@
 #                            FORWARD_DECL_MODULES_VAR fdecl_module_list)
 #
 #    # module2 -- dynamically linked
-#    include_directories({Boost_INCLUDE_DIRS})
+#    include_directories(${Boost_INCLUDE_DIRS})
```
henryiii, line 204: Nice! I was looking for an error, but missed this. :)

henryiii:
> I am currently discussing with the CMake team and will follow up shortly with an update.

That would be great. The "sphinxcontrib-cmakedomain" package seems to be abandoned, so we could ask for that name on PyPI if you wanted something simpler - we could even release to both names for a while. Also, I don't see a "sphinxcontrib-cmake" name, I'm guessing @slurps-mad-rips (and the other package's author) considered that, so maybe there's a reason it's best to avoid just "sphinxcontrib-cmake" here. (?) Anyway, there are existing users, so keeping the same name and just documenting it (or simply having it in the scikit-build org) is fine for discoverability.
 
> @henryiii Do you think including the update of the `moderncmakedomain` as part of the `cmake-python-distributions` release process something reasonable ?
> 
> In a nutshell, this would consist in copying the `cmake.py` from CMake code base, while would trigger the continuous integration and automatic update to PyPI (based on GitHub Action) that we would be adding to the project.

Yes, something like that would be ideal. I'd want it mostly automated, though it could be a separate button for now. Ideally, it would be nice if the CMake releases would trigger the PR creation, which I think from the links you sent earlier is likely possible. Eventually auto-triggering was in the plan when working on the nox tooling.

slurps-mad-rips:
> 
> 
> > I am currently discussing with the CMake team and will follow up shortly with an update.
> 
> That would be great. The "sphinxcontrib-cmakedomain" package seems to be abandoned, so we could ask for that name on PyPI if you wanted something simpler - we could even release to both names for a while. Also, I don't see a "sphinxcontrib-cmake" name, I'm guessing @slurps-mad-rips (and the other package's author) considered that, so maybe there's a reason it's best to avoid just "sphinxcontrib-cmake" here. (?) 

It's considered "well mannered" for sphinx plugins to add `domain` for plugins that add a domain and to not have it present for anything that is not related to a domain. Its not enforced, but I was following de facto best practices that the sphinx community have implicitly agreed upon 🙂

henryiii:
> It's considered "well mannered" for sphinx plugins to add domain

Thats fine, then, not that familiar with sphinx conventions. So it might be worth trying to get the `sphinxcontrib-cmakedomain` name, though, to eventually move it to that? Most maintainers if still around are happy to give up a name if there's a clearly better maintained version of it, and if the maintainer is unreachable, we can request the name from PyPI (it's been dead for 8 years). I can probably set up the setup.py to produce the "other" name on request, so we can dual-release for a while. I do like the term [Modern CMake](https://cliutils.gitlab.io/modern-cmake/), so I don't mind the name moderncmakedomain either :), just thinking `sphinxcontrib-cmakedomain` might be a bit more official and discoverable.

slurps-mad-rips:
They're definitely different in what they do and how they are declared and it could possibly break other folks stuff. That said PyPI didn't have squatting/maintainership policy back when I created the plugin. We'd have to contact the author however and I believe the PyPI folks can help in that regard if necessary, though it can be months before you receive a response of any kind 😅

henryiii:
I don't think anyone is using code from 8 years ago, I'm sure it doesn't support Python 3, etc. And sphinxcontrib-moderncmakedomain has 300x more downloads - probably pretty much only mirrors for the old package. I don't think it should be a blocker, but it might be nice to take that name too and maybe slowly transition if we can. Anyway, let's see what @jcfr and the KitWare team comes back with. :)

slurps-mad-rips:
It appears you would be [right](https://libraries.io/pypi/sphinxcontrib-cmakedomain). It seems there are no *public* dependencies on `cmakedomain` 🙂

henryiii:
I also checked https://pypistats.org/packages/sphinxcontrib-cmakedomain and compared it with https://pypistats.org/packages/sphinxcontrib-moderncmakedomain

henryiii:
And the readme recommends `pip install sphinxcontrib-cmakedomain==0.2`, which would be safe.

jcfr:
:100: 

> Anyway, let's see what @jcfr and the KitWare team comes back with. :)

After sharing our plan with the CMake team, transferring the repo to the `scikit-build` organization is a sensible path forward. 

The authoritative module would still live within the CMake code base and we could update the release process of `cmake-python-distributions`to include the update of the sphinx add-on as a necessary step.

Basically copying the module from the CMake code base, which will trigger CI and automatic update to PyPI associated with `sphinxcontrib-moderncmakedomain` project.

How does that sound ?

jcfr:
> I'd like to get the PyPI ownership swapped out before I transfer the github repository
> as well as who to add as an admin/owner for the pypi package

You could add:
* [henryiii](https://pypi.org/user/henryiii/)
* [jcfr](https://pypi.org/user/jcfr/)
* [kwrobot](https://pypi.org/user/kwrobot/)


> would be easier to then transfer it to scikit-build

As described in the guide [Repository transfers and organizations](https://docs.github.com/en/github/administering-a-repository/managing-repository-settings/transferring-a-repository#repository-transfers-and-organizations)


I just added you to the scikit-build organization, you should now be able to transfer the repository.

image

image

slurps-mad-rips:
> In the process, I also suggest to rename the project `sphinx-moderncmakedomain`

I absolutely missed this, someone else will have to take care of it, sorry. 😔 

The project has been transferred along with an updated `History` section in the README to the scikit-build organization.

PyPI ownership will be added shortly.

slurps-mad-rips:
PyPI invites have been sent out. Once someone has accepted it, I'll remove myself as an owner 🙂

PR #577, opened 2021-08-06
Update release process to include GitHub release creation and sending of annoucements on mailing list

jcfr:


CHANGES.rst
```diff
@@ -56,7 +56,10 @@ Documentation
 
 * Update `Conda: Step-by-step` release guide available in :doc:`/make_a_release` section.
 
-* Update links to CMake documentation pages in :doc:/`generators`. Thanks :user:`Eothred` for the contributions in :pr:`508`.
+* Update links to CMake documentation pages in :doc:`/generators`. Thanks :user:`Eothred` for the contributions in :pr:`508`.
+
+* Update :doc:`/make_a_release` to include creation of the GitHub release and sending of announcements on
+  the `scikit-build mailing list <https://groups.google.com/g/scikit-build>`_.
```
henryiii, line 62: I'd actually very lightly recommend not mentioning this. It's only valid for maintainers, so makes the changelog less readable for users. One can always look at the git log if they are a maintainer / very serious. I'm very fond of a great changelog, but listing too much in a changelog is only slightly better than not enough.

henryiii (approved):
See suggestion, though I'm fine either way.

PR #578, opened 2021-08-06
Fix misspellings and add codespell GitHub action

jcfr:


henryiii:
I've pushed a comment moving the check to pre-commit instead of a custom Actions job. This is faster, integrated with our existing and future checks, and is not tied to actions - you can just run "pre-commit run -a" locally and check the spelling before committing (or pre-commit install to add a githook that always checks it on commenting). It also gets pinned and updated regularly.

See https://scikit-hep.org/developer/style#spelling-extra

Do we want to keep the config in the pre-commit configuration, or should we add a codespell config file? I'm happy with it inline like this, but I haven't thought about it much, as I didn't know it accepted one until now. You can even add the config in setup.cfg - that's actually _really_ tempting. Sadly, no pyproject.toml support.

henryiii:
FYI, I'll be working on expanding the checks after we release 0.12, there are lots of useful ones.

henryiii (approved):
Moving to having the config in setup.cfg is optional, either way is fine with me.

jcfr:
>you can just run "pre-commit run -a" locally

This is great

jcfr:
> Do we want to keep the config in the pre-commit configuration, or should we add a codespell config file? 

Since we can run the pre-commit locally, keeping the config in  the `.pre-commit-config.yaml` is simple enough for now.

> You can even add the config in setup.cfg - that's actually really tempting. Sadly, no pyproject.toml support.

Considering that in the future we will likely drop use of `setup.cfg` ... and there are no `pyproject.toml` support, not worth it at that stage. We could still add `.codespellrc` to support running the tool in a standalone fashion.

I also started to create an issue in the codespell repo but realized that supporting loading config from `pyproject.toml` has some implication and it not as trivial as one would think (at least based on discussion adding such support in flake8). For context:
* https://gitlab.com/erikkemperman/flake8/-/commit/d720e01c5316727492737871858b78366fdbe8b9
* https://github.com/PyCQA/flake8/issues/234

henryiii:
Anthony is the main reason. :) Most of the issues he has with it there have gone away. TOML has reached 1.0 now, `tomli` has become the "standard" TOML implementation (over `toml`) and reads TOML 1.0 and it itself 1.0 - there might even be an effort to get it into the stdlib, and pip is planning to default to a modern build even if pyproject.toml is missing (end of year at the earliest, though, and might be delayed). Also, so far, everyone except black supports a "classic" setup.cfg or custom named file as well.

Most projects support it now, flake8 is the standout (MyPy just added it).

PR #581, opened 2021-08-11
chore: post-release and deployment update

henryiii:


.github/workflows/cd.yml
```diff
@@ -1,22 +1,39 @@
...
 jobs:
-  publish:
-    name: Publish
+  dist:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-      with:
-        fetch-depth: 0
+    - uses: actions/checkout@v1
```
henryiii, line 13:
```suggestion
    - uses: actions/checkout@v2
       with:
         fetch-depth: 0
```

.github/workflows/cd.yml
```diff
@@ -17,6 +17,25 @@ jobs:
...
+
+    steps:
+    - uses: actions/download-artifact@v2
+      with:
+        name: artifact
+        path: dist
+
     - uses: pypa/gh-action-pypi-publish@v1.4.2
       with:
         password: ${{ secrets.pypa_password }}
```
henryiii, line 41:
```suggestion
        password: ${{ secrets.pypi_password }}
```

.github/workflows/cd.yml
```diff
@@ -17,6 +17,25 @@ jobs:
...
+        path: dist/*
+
+    - name: Check metadata
+      run: pipx run twine check dist/*
+
+
+  publish:
+    needs: [dist]
+    runs-on: ubuntu-latest
+    if: github.event_name == 'release' && github.event.action == 'published'
```
jcfr, line 31: Thanks for implementing this, it is great that now have a manual way of triggering the CD workflow along with upload of the packages without systematically publishing to PyPI.

Issue #579, opened 2021-08-06
Potential bug: presence of a subpackage lets editable install point to 'project root dir' instead of 'src'


maichmueller:
Hi all,

I have noticed a strange behaviour while building a package of mine, which may be due to my configuration or a bug.
I have compiled a test project with a blank python module that I am compiling using scikit-build. This project follows the 'src' layout, which comes down to this:
```
├── CMakeLists.txt
├── pyproject.toml
├── setup.py
└── src
    └── pkg
        ├── core
        │   └── __init__.py
        ├── __init__.py
        ├── module.cpp
        └── some_funcs.py

```

The weird part is this: The mere presence of the subpackage `core` lets `pip install -e .` point to the root directory of this project `.` instead of `src`, the actual package directory:
> Creating /home/michael/anaconda3/envs/testpkg/lib/python3.8/site-packages/pkg.egg-link (link to .)

 This causes the editable install to be useless, as one would need to call `import src.pkg` in python to import this. 
It points correctly to `src` when `core` is removed from the package, or when one does not use scikit-build (i.e. uses `setuptools.setup` and does not compile the extension:
> Creating /home/michael/anaconda3/envs/testpkg/lib/python3.8/site-packages/pkg.egg-link (link to src)

In order to reproduce this behaviour I have put together a minimal project that does this (https://github.com/maichmueller/scikit_build_issue). The steps to reproduce are:

```
git clone https://github.com/maichmueller/scikit_build_issue.git
cd scikit_build_issue
pip install -e .
```

In order to see it pointing to `src`, either do:
1) delete `core` from `src/pkg` or
2) uncomment the last lines in setup.py to use setuptools' `setup` function (and comment out skbuild's `setup` call).

Is this an error on my side due to faulty configuration? The non working setup py reads the following setup call:
```
setup(
    name="pkg",
    version="0.0.1",
    packages=find_packages(where="src"),
    package_dir={"": "src"},
    cmake_install_dir="src/pkg",
    cmake_args=[f"-Dpybind11_SEARCH_PATH={pybind11_path}"],
    zip_safe=False,
    include_package_data=True,
)
```

Help or feedback to what might be happening here would be appreciated! 

Thanks,
Michael

maichmueller:
Just noticed my issue is a duplicate of #546

maichmueller:
So I have been digging a little into the setuptools process and noticed where the editable install breaks with skbuild:

[...]

Issue #546, opened 2021-06-29
Broken editable install with sub-package


jwittbrodt:
Hi,

I have been getting ModuleNotFoundErrors in my project when installing with `pip install -e .` while everything works fine for `pip install .`. I then tried a pure python toy project and found that it works with `setuptools.setup` but fails with `skbuild.setup` and seems to be related to a sub-package.

The toy project has the following trivial layout

📦project
 ┣ 📂src
 ┃ ┣ 📂example
 ┃ ┃ ┣ 📂tools
 ┃ ┃ ┃ ┗ 📜__init__.py
 ┃ ┃ ┗ 📜__init__.py
 ┗ 📜setup.py

and the `setup.py` I am using is simply
```python
# failing case
from skbuild import setup

setup(
    name="example_package",
    packages=["example", "example.tools"],
    package_dir={"": "src"}
)
```
If I then do `pip install -e .` I get a ModuleNotFoundError (while `pip install .` works without issues).

Now if I remove the sub-package, i.e. change the setup.py to
```python
# working case 1
from skbuild import setup

setup(
    name="example_package",
    packages=["example"],
    package_dir={"": "src"}
)
```
it works without problems both in editable mode and without.

Similarly, if I keep the sub-package but use setuptools instead of skbuild, i.e.
```python
# working case 2
from setuptools import setup

setup(
    name="example_package",
    packages=["example", "example.tools"],
    package_dir={"": "src"}
)
```
both ways work as well.

Looking at the `.egg-link` generated in those cases, it seems like for the two working cases it correctly points to `project/src` and also generates the `.egg-info` directory there. However, in the failing case it points only to `project` and generates `.egg-info` directories both in `project` and in `project/src`.

I am using
 - python 3.9.5
 - scikit-build 0.11.1
 - setuptools 49.2.1
on an up-to-data arch linux machine.

I don't really know enough about the setuptools process to dig any further, so any help would be appreciated.
Cheers,
Jonas

henryiii:
How did you generate the little tree above? Was it manual? 😎

jwittbrodt:
@henryiii I used this vscode extension: https://marketplace.visualstudio.com/items?itemName=Shinotatwu-DS.file-tree-generator

ruestefa:
I've just experienced the exact same problem, I'm really glad I found this issue so I know the problem's not with my setup. Are there any news regarding a potential fix?

[...]

PR #582, opened 2021-08-11
Potential fix for broken editable install

maichmueller:
this small change in setuptools_wraps I believe fixes the broken editable install links due to the `src` directory of the package being set incorrectly (#579 and #546 ). 
I am unsure if this change holds other consequences for the scikit-build process, so this PR could also just be seen as a hint towards what direction of change would fix the issue if one of the maintainers wants to take this on themselves.

### DIFF
--- a/tests/samples/issue-546-editable-subproject/src/pkg/__init__.py
+++ b/tests/samples/issue-546-editable-subproject/src/pkg/__init__.py
@@ -1,0 +1,3 @@
+from ._pkg import LogicCPP
+
+__all__ = ("LogicCPP",)
--- a/skbuild/setuptools_wrap.py
+++ b/skbuild/setuptools_wrap.py
@@ -720,14 +720,19 @@ def setup(*args, **kw):  # noqa: C901
             _copy_file(data_file, dest_data_file, hide_listing)
 
     kw["package_data"] = package_data
-    kw["package_dir"] = {
-        package: (
-            os.path.join(CMAKE_INSTALL_DIR(), prefix)
-            if os.path.exists(os.path.join(CMAKE_INSTALL_DIR(), prefix))
-            else prefix
-        )
-        for prefix, package in package_prefixes
-    }
+
+    # Update the entries of the package_dir kw without removing main package entry
+    package_dir.update(
+        {
+            package: (
+                os.path.join(CMAKE_INSTALL_DIR(), prefix)
+                if os.path.exists(os.path.join(CMAKE_INSTALL_DIR(), prefix))
+                else prefix
+            )
+            for prefix, package in package_prefixes
+        }
+    )
+    kw["package_dir"] = package_dir
 
     kw["scripts"] = [
         os.path.join(CMAKE_INSTALL_DIR(), script) if mask else script for script, mask in new_scripts.items()
--- a/tests/samples/issue-546-editable-subproject/CMakeLists.txt
+++ b/tests/samples/issue-546-editable-subproject/CMakeLists.txt
@@ -1,0 +1,14 @@
+cmake_minimum_required(VERSION 3.15)
+
+project(testpkg VERSION 0.1 LANGUAGES CXX)
+
+set(PYBIND11_FINDPYTHON TRUE)
+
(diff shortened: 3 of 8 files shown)


henryiii:
This looks like it would be pretty easy to test. We could add a test for #546 pretty easily, see if it breaks, then try with this fix.

tests/samples/issue-546-editable-subproject/src/pkg/__init__.py
```diff
@@ -0,0 +1 @@
+from ._pkg import *   # noqa: F403,F401
```
henryiii, line 1:
```suggestion
from ._core import logic
```
?

maichmueller: I'm not sure I'm understand the idea of this change. 
`_core` has been renamed to `subpkg` in the meantime to make clear that this is just a standalone sub package. So if any we should add `from ._subpkg import *`. But there is nothing to export in `subpkg` so this wouldn't change much I guess. 
The extension `_pkg` is installed in the module folder `pkg` and exposes currently only the name `LogicCPP`. I can change the star import to that if you want, but I think the linter would still complain without the ignore codes. 
In general, feel free to change the test case however you see fit. This case is simply the one I managed to make skbuild fail the editable install with.

maichmueller: I have instead renamed the pybind11 module to _pkg as well (since I renamed the test module to pkg and subpkg with the 'internal' lib being _pkg). Would changing the star import be important?

jwittbrodt:
I can confirm that the test project here reproduces the issue for me with the latest skbuild release (this scenario of a pybind11 package with a pure python sub-package is in fact exactly the use case where I first encountered the issue).

maichmueller:
thanks for confirming @jwittbrodt !

henryiii:
Looks like just the editable install is failing (at least on the one that has finished so far) :)

I've fixed the actions so it should run when you make an edit without me having to click approve.

henryiii:
Curious, it seems to be that windows passes the non-editable install check, but linux does not. Are you on Windows too?

maichmueller:
Thanks for lifting the restrictions! That should make testing faster.

I am on a Windows machine but I am running the project in a ubuntu docker container (the test case with the dockerfile I am using is here: https://github.com/maichmueller/scikit_build_issue/blob/master/Dockerfile). 

Which workflow run that succeded do you mean? It seems to me that all have failed so far?

PS

henryiii:
All workflows fail, because the editable install fails. I'm mostly looking at the non-editable install, which I'd expect to work. It seems to work on Windows, but not Linux. This might be due to Python 2, which only runs on non-Windows.

Still getting `No module named 'pkg'` (on both now that I've renabled the fix in this PR).

maichmueller:
Have you tested the case locally to see if it installs the pkg correctly there (for both types of install)?

maichmueller:
So I have tested my fix now locally on my Ubuntu 20.04 (py38) station and saw it fix the issue. 
I have then tested this PR's fork of skbuild by running pytest on the relevant skbuild tests on this machine and saw the classic one **not** fail, but the editable one **did**. I can't really dig into what is happening during the tests, because I have no idea what is happening during the test setup. E.g. I can't..
1. hook up the debugger onto pytest (getting a warning about sys.settrace having been modified)
2. any changes to the skbuild source code are not executed in pytest (e.g. tried immediately returning out of setuptools_wraps.setup and that never is executed)
3. Because of 2. can't print any variable values or other info.

How can I modify the code in a way that takes effect for the testcases? Without that, I can't quite dig any further into differences between my local setup and the test setup.

henryiii:
I'll try to help soon-ish, but https://scikit-hep.org/developer/pytest#running-pytest describes how to debug with pytest. You can use `--trace` to start a debugger at the beginning, or `--pdb` to start the debugger when it fails. You can use `--lf` to start at the last failed test (good for `--trace`). You can pass arguments to pytest through nox with `--`.

I think there was a caching issue that I've just fixed with the test.

henryiii:
The regular install works with/without, and the editable install does not work with or without the patch. :(

PR #585, opened 2021-08-23
tests: add editable test

henryiii:
@jwittbrodt or @maichmueller, any idea how to make this new test fail?

### DIFF
--- a/tests/test_issue546_editable_subproject.py
+++ b/tests/test_issue546_editable_subproject.py
@@ -5,4 +5,4 @@ DIR = os.path.abspath(os.path.dirname(__file__))
 
 def test_editable_subproject(virtualenv):
     virtualenv.run("pip install -e {DIR}/samples/issue-546-editable-subproject".format(DIR=DIR))
-    virtualenv.run("python -c 'import example.tools'")
+    virtualenv.run('python -c "import example.tools"')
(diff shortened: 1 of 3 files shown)


henryiii:
I would hope this new fails, and then https://github.com/scikit-build/scikit-build/pull/582 would be required to fix it.

tests/test_issue546_editable_subproject.py
```diff
@@ -0,0 +1,8 @@
+import os
+
+DIR = os.path.abspath(os.path.dirname(__file__))
+
+
+def test_editable_subproject(virtualenv):
+    virtualenv.run("pip install -e {DIR}/samples/issue-546-editable-subproject".format(DIR=DIR))
+    virtualenv.run("python -c 'import example.tools'")
```
henryiii, line 8:
```suggestion
    virtualenv.run('python -c "import example.tools"')
```

maichmueller:
@henryiii I could not get scikit-build to fail the editable install with the minimal python only package example from #546 . However, it did fail for a minimal cpp extension (pybind11) with subpackages. I have added that test case to my PR and I am testing whether the current skbuild version fails on that in the workflows. If it does, then I will re-add my fix and see if that resolves it.

jwittbrodt:
Thanks a lot for looking into this. I also can't get my pure python example to fail anymore, seems some new version somewhere fixed that issue. But as @maichmueller said, it still fails for me if a pybind11 extension is involved.

henryiii:

PR #597, opened 2021-10-26
Accommodate packages in separate source directory

jcrudy:
I couldn't figure out how to get my package to install correctly when it's in a source directory other than the one containing setup.py.  This change does what I need, though perhaps it's not the best way to go about it.


jcrudy:
This seems to be unneeded, as the problem is solved by the `cmake_install_dir` argument.  Closing.

PR #606, opened 2021-11-10
feat: support CMAKE_ARGS

### LINKED ISSUE #588: Support CMAKE_ARGS
Some tools (conda) set CMAKE_ARGS; ideally I think that should be automatically supported. It's also really nice to have environment variable methods to pass things through, since modern build systems do not have direct access to the setup.py call.

I'd propose something like:

```python
if "CMAKE_ARGS" in os.environ:
    cmake_args += [arg for arg in os.environ["CMAKE_ARGS"].split(" ") if arg]
```

Might need to handle quotes. See https://github.com/tdegeus/QPot/pull/36 and https://github.com/pybind/cmake_example/pull/58

henryiii:
Closes #588, would make supporting conda-forge easier, less workarounds like the ones in that issue and https://github.com/conda-forge/correctionlib-feedstock/blob/1d49847e4ac8e69967142146e8cbac81173ee3d3/recipe/setup.patch required.

If this looks like a good idea I can add tests and such.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -514,6 +514,21 @@ If a file is added to the CMake build system by updating one of the ``CMakeLists
 will not explicitly reconfigure the project. Instead, the generated build-system will automatically
 detect the change and reconfigure the project after :meth:`skbuild.cmaker.CMaker.make` is called.
 
+
+Environment variable configuration
+----------------------------------
+
+Scikit-build support environment variables to configure some options. These are:
+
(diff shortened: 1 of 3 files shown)


henryiii:
Actually, it looks like `SKBUILD_CONFIGURE_OPTIONS` does this already - maybe the better solution here is to get conda to support that.

henryiii:
@jcfr, @thewtex thoughts? This would also be useful for things like Pyodide, where they are mostly imitating conda-forge, so setting `CMAKE_ARGS` would be more natural for them, too.

PR #607, opened 2021-11-10
fix: avoid forcing the macOS version

### LINKED ISSUE #596: bug: compute correct deployment target
See https://github.com/pypa/cibuildwheel/issues/818. This was closed as won't-fix because it's Scikit-build's fault, not cibuildwheel's.

`MACOS_DEPLOYMENT_TARGET` should override the macOS version in `_PYTHON_HOST_PLATFORM` if it's higher, just like in setuptools.

henryiii:
This fixes #596 by removing the custom logic that is lowering the macOS version incorrectly. This was originally there because Python defaults to whatever it was built with, and 10.6 was too low for modern macOS. However, all normal Pythons (everything except the dead 3.5) have 10.9 or better installers now, and this is causing the (correct) calculation of the version to be lost. Users should always set `MACOS_DEPLOYMENT_TARGET`, and most wheel building tools will do this for you just in case.

Draft until I'm sure it works or can add tests.

### DIFF
--- a/skbuild/command/bdist_wheel.py
+++ b/skbuild/command/bdist_wheel.py
@@ -1,9 +1,6 @@
 """This module defines custom implementation of ``bdist_wheel`` setuptools
 command."""
 
-import os
-import sys
-
 try:
     from wheel.wheelfile import WheelFile
     _USE_WHEELFILE = True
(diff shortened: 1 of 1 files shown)


henryiii:
Before:

```console
$ MACOSX_DEPLOYMENT_TARGET=10.13 _PYTHON_HOST_PLATFORM=macosx-10.11-universal2 pipx run build --wheel
...
Successfully built scikit_build_example-0.0.1-cp310-cp310-macosx_10_11_universal2.whl
```

After:

```console
$ MACOSX_DEPLOYMENT_TARGET=10.13 _PYTHON_HOST_PLATFORM=macosx-10.11-universal2 pipx run build --wheel
...
Successfully built scikit_build_example-0.0.1-cp310-cp310-macosx_10_13_universal2.whl
```

I'm happy with it! :)

henryiii:
Ouch, this bites PyPy because it defaults to 10.7, instead of an actual proper minimum (10.9+). @mattip I'm sure I've mentioned this before - do you know if there's been any progress on that (or where we discussed it before)? Setuptools (and therefore Scikit-Build) require `MACOSX_DEPLOYMENT_TARGET` (cibuildwheel sets it by default) to build a PyPy extension on macOS, while CPython is fine because they support 10.9+ instead of 10.7+, so they get stdc++ instead of stdlibc++, which was removed some time ago.

I'm still in favor of this patch since it's exactly the same issue with exactly the same fix as vanilla setuptools, but it would be nice to fix it properly in PyPy. We could also consider setting `MACOSX_DEPLOYMENT_TARGET` if unset and PyPy is detected, which would be a less problematic workaround for now.

mattip:
This will be 10.9 for the upcoming release.

Issue #261, opened 2017-06-23
target_link_libraries_with_dynamic_lookup does not support  keyword signature


henryiii:
When I try to combine PyBind11 and Scikit-build I get:
```
    CMake Error at /usr/local/lib/python2.7/site-packages/skbuild/resources/cmake/targetLinkLibrariesWithDynamicLookup.cmake:577 (target_link_libraries):
      The keyword signature for target_link_libraries has already been used with
      the target "goofit".  All uses of target_link_libraries with a target must
      be either all-keyword or all-plain.

      The uses of the keyword signature are here:

       * extern/pybind11/tools/pybind11Tools.cmake:156 (target_link_libraries)
       * extern/pybind11/tools/pybind11Tools.cmake:103 (target_link_libraries)
       * python/CMakeLists.txt:49 (target_link_libraries)
```

henryiii:
This is a problem with the CMake module, but the issue is probably that PyBind does it's own setting up to prepare the `.so`, and I just need to figure out how to link to it (instead of using the skbuild extras in conjunction with the pybind11 module)`

jcfr:
xref #255

maxbachmann:
I did just run into the same issue with `target_link_libraries`
```
target_link_libraries(cpp_process_cdist PRIVATE Threads::Threads)
```
fails with `All uses of target_link_libraries with a target must be either all-keyword or all-plain.`, while
```
target_link_libraries(cpp_process_cdist Threads::Threads)
```
works

Issue #243, opened 2017-02-03
UseCython: Make use of modern CMake


jcfr:
Currently the only way to add to the Cython module search path for add_cython_target is to use the outdated include_directories CMake command. Ideally, we would be able to use target_include_directories or something similar. This may involve the creation of an actual CMake target.

See https://github.com/scikit-build/scikit-build/blob/master/skbuild/resources/cmake/UseCython.cmake

cmpute:
This feature is exactly what I need now! I think the generator expression in CMake would help this, like `set(prop "$<TARGET_PROPERTY:INCLUDE_DIRECTORIES>")` as said in [CMake documents](https://cmake.org/cmake/help/v3.9/manual/cmake-generator-expressions.7.html#id4). Also, I think the way you mentioned to add include directories to CMake should be added to the documentation!~

henryiii:
scikit-build (classic) after #1185 is now a thin wrapper around scikit-build-core's setuptools plugin. Closing all open issues; if this still applies, please reopen or open an issue on scikit-build-core. Thanks!

PR #611, opened 2021-11-24
Add option to use modern form of target_link_libraries

maxbachmann:
This adds the option SKBUILD_LINK_LIBRARIES_KEYWORD which can be set to PRIVATE,
PUBLIC, or INTERFACE or left empty (the default) to use the old form of
target_link_libraries.

A similar change was done in FindCuda.cmake: https://gitlab.kitware.com/cmake/cmake/-/commit/9f41bfd7b9e6e8a7545f741f872949d2ae801978
This fixes https://github.com/scikit-build/scikit-build/issues/261
This still requires documentation and I have to test it.

### DIFF
--- a/skbuild/resources/cmake/FindPythonExtensions.cmake
+++ b/skbuild/resources/cmake/FindPythonExtensions.cmake
@@ -434,7 +434,7 @@ endfunction()
 
 function(python_standalone_executable _target)
   include_directories(${PYTHON_INCLUDE_DIRS})
-  target_link_libraries(${_target} ${PYTHON_LIBRARIES})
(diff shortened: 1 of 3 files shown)


henryiii:
I'd be okay with this as a stopgap measure for now. Hopefully #601 will get accepted and we'll redesign the modules in the future. :)

maxbachmann:
At least for me this is not super important right now, so it would be fine to wait until the modules are redesigned. I really hope #601 gets accepted, since it has a ton of features/improvements I would love to see in scikit-build.
- proper support for PyPy
- proper support for Cython
- modern CMake support (#261, #243)
- defining everything in pyproject.toml without needing setup.py/setup.cfg

henryiii:
I think we should merge this in as a "hidden" feature for now, as it gives us additional workarounds & flexibility in the future.

PR #612, opened 2021-11-30
Add ARM64 as a new Windows platform

gaborkertesz-linaro:
Enable new Windows platform for CMakeVisualStudioIDEGenerator.

Tested with CMake Python distribution wheel generation on win-arm64.


henryiii (approved):
I'd be happy to take this as "experimental support for ARM" until we either get a Windows ARM box, or CI starts supporting it (which I've heard might be soon).

gaborkertesz-linaro:
> I'd be happy to take this as "experimental support for ARM" until we either get a Windows ARM box, or CI starts supporting it (which I've heard might be soon).

Thanks, really appreciated! If I can do anything to make it more smooth, just please let me know!

henryiii:
We are all very busy with #601 (due today), but after that, unless there are objections, I'd say we should merge.

Issue #618, opened 2022-01-03
Python library not found when using recent setuptools with pypy


rkaminsk:
It seems like there is a problem using pypy 7.3.7 together with scikit-build and setuptools version 64.2.0 because cmake is called with `-DPYTHON_LIBRARY=`. Using the setuptools version 59.4.0, everything works and a correct library path is passed. Does anyone know about this problem?

I originally reported the problem in the manylinux project pypa/manylinux#1246 where I also attached log files.

rkaminsk:
I looked a bit around and found the problem in `scikit-build/skbuild/cmaker.py` in function `get_python_library`. With newer setuputils versions, function `du_sysconfig.get_config_var('LIBDIR')` does not behave as before anymore:

```python
>>> import distutils.sysconfig as du_sysconfig
>>> import sysconfig
>>> du_sysconfig.get_config_var('LIBDIR')
'/build_dir/pypy-c-jit-linux-x86-64/build/bin'
>>> sysconfig.get_config_var('LIBDIR')
'/snap/pypy3/95/bin'
```

The above code ran on Ubunutu with pypy installed via snap but something similar happens when running in a manylinux container to build platform-specific wheels.

I think it would be relatively easy be fixed by adding both libdirs to the candidates cross-product. What do you think?

mlasch:
I looked a bit around on this topic, because I have also an issue in this area (#621). The `distutils` module is deprecated and will be removed from the standard library in Python 3.12, see [PEP632](https://www.python.org/dev/peps/pep-0632/). I guess there needs to be a solution based on `sysconfig` only.
The `sysconfig` module was introduced with Python 3.2 so backwards compatibility should be fairly well guaranteed.

rkaminsk:
> I looked a bit around on this topic, because I have also an issue in this area (#621). The `distutils` module is deprecated and will be removed from the standard library in Python 3.12, see [PEP632](https://www.python.org/dev/peps/pep-0632/). I guess there needs to be a solution based on `sysconfig` only. The `sysconfig` module was introduced with Python 3.2 so backwards compatibility should be fairly well guaranteed.

It would be relatively easy to make this future proof by conditionally importing the `distutils.sysconfig` module. (At least in the one file of the project I looked at.) We should probably wait for the project maintainers for some feedback, though.

mlasch:
> It would be relatively easy to make this future proof by conditionally importing the distutils.sysconfig module.

I think distutils is not even needed anymore.

[...]

PR #620, opened 2022-01-04
Improve detection of python library path

rkaminsk:
This patch considers the LIBDIR variable from both `distutils.sysconfig` and `sysconfig`. This is necessary to correctly detect the python library path when using pypy with recent setuputils versions.  (See #618.)


henryiii:
Any idea about the one failure?

rkaminsk:
> Any idea about the one failure?

No, clue. It does not make sense to me. I can spend some more time tomorrow, if you are interested in this PR.

henryiii:
Thanks, I would hope it would enable #624.

rkaminsk:
I'll debug it and come back to you.

rkaminsk:
I found the problem. I overlooked one case. Now it is working as intended.

PR #624, opened 2022-01-13
ci: test pypy

henryiii:
Adding PyPy. Two tests fail on macOS, but having the rest of the tests is better than not having them, and hopefully we can solve those issues eventually.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,6 +141,10 @@ jobs:
         echo "SKBUILD_TEST_FIND_VS2017_INSTALLATION_EXPECTED=1" >> $GITHUB_ENV
         echo "SKBUILD_TEST_FIND_VS2019_INSTALLATION_EXPECTED=1" >> $GITHUB_ENV
 
+    - name: Set min macOS version
+      run: |
+        echo "MACOS_DEPLOYMENT_TARGET=10.9" >> $GITHUB_ENV
+
     - name: Setup nox
       uses: excitedleigh/setup-nox@v2.0.0
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,11 +52,6 @@ jobs:
     - name: Setup nox
       uses: excitedleigh/setup-nox@v2.0.0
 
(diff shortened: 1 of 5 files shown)


henryiii:
1.

rkaminsk:
Since I was looking on a similar issue on Ubuntu, I also had a quick look here. It would be relatively easy to fix this on macos. If you run `pypy -m sysconfig`, you see the following:

    LDLIBRARY = "libpypy3-c.dylib"
    LIBDIR = "/Users/runner/hostedtoolcache/PyPy/3.7.12/x64/bin"

The concatenation of both should be the path to the library. I think I can provide a PR to fix this, too.

henryiii:
Fantastic, thanks! @mattip, possibly this is also fixable in PyPy? (WITH_DYLD is not set on PyPy macOS, may be more obvious in the linked PR).

mattip:
`sysconfig.get_config_var('WITH_DYLD')` is [1 on macOS](https://github.com/python/cpython/blob/d75a51bea3c2442f81d38ff850b81b8b7f3330f0/configure#L10130) and None everywhere else. Is that the best way to detect macOS?

mattip:
For instance, I think `sysconfig.get_config_var('MULTIARCH') == 'darwin'` is possible

henryiii:
We add `.dylib` if `WITH_DYLD` is set. Given it looks like it is always set on macOS and never otherwise, likely could just base it on the OS, but I think that's what CPython provides this for.

mattip:
With that, it is easy enough to add `WITH_DYLD`, I just wish sysconfig had a spec so we could say at some point it is complete.

mattip:
Done, it will be part of the next release. Of course older PyPys will not have the value.

henryiii:
Thanks! I'm hoping (if https://iscinumpy.dev/post/scikit-build-proposal/ is accepted) that I'll be rewriting scikit-build into scikit-build-core at the end of the year, and if so, I can probably avoid this workaround and just support the latest PyPy. :)

> I just wish sysconfig had a spec

That would be nice! Might be worth asking about, such as on the PyPA discord server or discuss.python.org?

PR #626, opened 2022-01-14
Use follow_symlinks=False in copyfile/copymode invocations

### LINKED ISSUE #625: Support SOVERSION-ed libraries
On Linux, for projects that set `VERSION`/`SOVERSION` properties on the dynamic library they install CMake generates symbolic links symbolic links for least specified version to the most specified one:

```
(work) [16:53:47 vmbox dpctl]$ ls -l _skbuild/linux-x86_64-3.9/cmake-install/dpctl/libDP*
_skbuild/linux-x86_64-3.9/cmake-install/dpctl/libDPCTLSyclInterface.so -> libDPCTLSyclInterface.so.0
_skbuild/linux-x86_64-3.9/cmake-install/dpctl/libDPCTLSyclInterface.so.0 -> libDPCTLSyclInterface.so.0.12
_skbuild/linux-x86_64-3.9/cmake-install/dpctl/libDPCTLSyclInterface.so.0.12
```

scikit-build copies these symbolic links as files, because it uses `shutils.copyfile` with `follow_symlinks=True` default (see [setuptools_wrap.py#L855-L856](https://github.com/scikit-build/scikit-build/blob/master/skbuild/setuptools_wrap.py#L855-L856)). Each file copied into Python layout is a regular file, and this interferes with proper working of the library.

I have fixed this issue locally, by modifying these 2 lines to 

```python
    copyfile(src_file, dest_file, follow_symlinks=False)
    copymode(src_file, dest_file, follow_symlinks=False)
```

oleksandr-pavlyk:
Closes #625

Use `follow_symlinks=False` non-default keyword when calling `shutil.copyfile` and `shutil.copymode` that is used for copying files linked in `cmake_manifest` to setuptools destination.

### DIFF
--- a/skbuild/setuptools_wrap.py
+++ b/skbuild/setuptools_wrap.py
@@ -834,6 +834,18 @@ def _classify_file(path, package_data, package_prefixes,
     file_set.add(os.path.join(CMAKE_INSTALL_DIR(), path))
 
 
+if sys.version_info.major < 3:
+    def _copy_file_and_mode(src_file, dest_file):
+        """Copy file and its mode"""
+        copyfile(src_file, dest_file)
+        copymode(src_file, dest_file)
+else:
+    def _copy_file_and_mode(src_file, dest_file):
+        """Copy file and its mode, preserving symbolic links"""
+        copyfile(src_file, dest_file, follow_symlinks=False)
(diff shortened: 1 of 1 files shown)


henryiii:

PR #630, opened 2022-01-23
Detect python library on macos + pypy

rkaminsk:
This improves on #624 reenabling the test for the python library which is found correctly now. I did not know how to fix the inplace build test, though.


henryiii:
Thanks! In place builds are borderline broken already, so a little extra issue there isn't too bad at the moment. We need to fix rebuilding in PEP 517 probably first - which means adding PEP 517 tests...

PR #631, opened 2022-01-25
fix: pull ninja path if module available

### LINKED ISSUE #628: CMake fails to find ninja on second build from pip when installed via build-requires
When building a project a second time using `pip install -e .`, cmake will fail to find ninja with the following error:
```
Failed to build whatever:
     'C:/Users/runneradmin/AppData/Local/Temp/pip-build-env-m4knfc7v/overlay/Scripts/ninja.exe' '--version'
    
      failed with:
    
       The system cannot find the file specified
```

This is caused by pip removing its temporary `pip-build-env` directory between the builds, and cmake not trying to look for ninja again despite it not existing.

OS: Windows 10.0.19044.0

henryiii:
Fixes #628.

~~Builds on / requires #634; must be able to force the version in order to get the tests to pull the local scikit_build wheel.~~

About 95% of the work was in adding the test, the fix was pretty easy. :)

### DIFF
--- a/skbuild/cmaker.py
+++ b/skbuild/cmaker.py
@@ -210,14 +210,6 @@ class CMaker(object):
         if cli_generator_name is not None:
             generator_name = cli_generator_name
 
-        ninja_executable_path = None
-        if (generator_name is None or generator_name == "Ninja"):
-            try:
-                import ninja
-                ninja_executable_path = os.path.join(ninja.BIN_DIR, "ninja")
-            except ImportError:
-                pass
-
         # if arch is provided on command line, use it
         clargs, cli_arch = pop_arg('-A', clargs)
 
@@ -225,6 +217,14 @@ class CMaker(object):
             generator_name, skip_generator_test=skip_generator_test,
             cmake_executable=self.cmake_executable, cmake_args=clargs,
             languages=languages, cleanup=cleanup, architecture=cli_arch)
+
+        ninja_executable_path = None
+        if generator.name == "Ninja":
+            try:
+                import ninja
+                ninja_executable_path = os.path.join(ninja.BIN_DIR, "ninja")
+            except ImportError:
+                pass
 
         if not os.path.exists(CMAKE_BUILD_DIR()):
             os.makedirs(CMAKE_BUILD_DIR())
--- a/skbuild/cmaker.py
+++ b/skbuild/cmaker.py
@@ -133,20 +133,24 @@ class CMaker(object):
         self.cmake_version = get_cmake_version(self.cmake_executable)
         self.platform = get_platform()
 
-    def get_cached_generator_name(self):
-        """Reads and returns the cached generator from the :func:`skbuild.constants.CMAKE_BUILD_DIR()`:.
-        Returns None if not found.
-        """
+    def get_cached(self, item):
+        """Return the cached value if possible, otherwise return None"""
+        item = '{}:'.format(item)
         try:
-            cmake_generator = 'CMAKE_GENERATOR:INTERNAL='
             with open(os.path.join(CMAKE_BUILD_DIR(), 'CMakeCache.txt')) as fp:
                 for line in fp:
-                    if line.startswith(cmake_generator):
-                        return line[len(cmake_generator):].strip()
+                    if line.startswith(item):
+                        return line.split("=", 1)[-1].strip()
         except (OSError, IOError):
             pass
 
         return None
+
+    def get_cached_generator_name(self):
+        """Reads and returns the cached generator from the :func:`skbuild.constants.CMAKE_BUILD_DIR()`:.
+        Returns None if not found.
+        """
+        return self.get_cached('CMAKE_GENERATOR')
 
     def get_cached_generator_env(self):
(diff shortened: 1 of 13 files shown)


skbuild/cmaker.py
```diff
@@ -133,21 +133,25 @@ def __init__(self, cmake_executable=CMAKE_DEFAULT_EXECUTABLE):
         self.cmake_version = get_cmake_version(self.cmake_executable)
         self.platform = get_platform()
 
-    def get_cached_generator_name(self):
-        """Reads and returns the cached generator from the :func:`skbuild.constants.CMAKE_BUILD_DIR()`:.
-        Returns None if not found.
-        """
+    def get_cached(self, item):
+        """Return the cached value if possible, otherwise return None"""
```
jcfr, line 137:
```suggestion
        """If possible, return the cached value from the :func:`skbuild.constants.CMAKE_BUILD_DIR()`, otherwise return None"""
```

jcfr: I changed `If possible` by `If set`

skbuild/cmaker.py
```diff
@@ -133,21 +133,25 @@ def __init__(self, cmake_executable=CMAKE_DEFAULT_EXECUTABLE):
         self.cmake_version = get_cmake_version(self.cmake_executable)
         self.platform = get_platform()
 
-    def get_cached_generator_name(self):
-        """Reads and returns the cached generator from the :func:`skbuild.constants.CMAKE_BUILD_DIR()`:.
-        Returns None if not found.
-        """
+    def get_cached(self, item):
```
jcfr, line 136:
```suggestion
    def get_cached(self, variable_name):
```

jcfr: Would also have to change `item` to `variable_name` in the code below

skbuild/cmaker.py
henryiii: Thanks! Working on getting a test working int this. Not trivial to get the local scikit-build into the PEP 518 requirements.

skbuild/cmaker.py
```diff
@@ -206,6 +210,14 @@ def configure(self, clargs=(), generator_name=None, skip_generator_test=False,
         if cli_generator_name is not None:
             generator_name = cli_generator_name
 
+        ninja_path = None
```
jcfr, line 213:
```suggestion
       ninja_executable_path = None
```

skbuild/cmaker.py
```diff
@@ -133,21 +133,25 @@ def __init__(self, cmake_executable=CMAKE_DEFAULT_EXECUTABLE):
...
+    def get_cached(self, item):
+        """Return the cached value if possible, otherwise return None"""
+        item = '{}:'.format(item)
         try:
-            cmake_generator = 'CMAKE_GENERATOR:INTERNAL='
             with open(os.path.join(CMAKE_BUILD_DIR(), 'CMakeCache.txt')) as fp:
                 for line in fp:
-                    if line.startswith(cmake_generator):
-                        return line[len(cmake_generator):].strip()
+                    if line.startswith(item):
```
henryiii, line 142:
```suggestion
                    if line.startswith(variable_name):
```

skbuild/cmaker.py
```diff
@@ -206,6 +210,14 @@ def configure(self, clargs=(), generator_name=None, skip_generator_test=False,
         if cli_generator_name is not None:
             generator_name = cli_generator_name
 
+        ninja_path = None
+        if (generator_name is None or generator_name == "Ninja"):
+            try:
+                import ninja
+                ninja_path = os.path.join(ninja.BIN_DIR, "ninja")
```
henryiii, line 217:
```suggestion
                ninja_executable_path = os.path.join(ninja.BIN_DIR, "ninja")
```

skbuild/cmaker.py
```diff
@@ -251,6 +263,8 @@ def configure(self, clargs=(), generator_name=None, skip_generator_test=False,
             cmd.extend(['-T', generator.toolset])
         if generator.architecture:
             cmd.extend(['-A', generator.architecture])
+        if ninja_path is not None:
+            cmd.append('-DCMAKE_MAKE_PROGRAM:FILEPATH=' + ninja_path)
```
henryiii, line 267:
```suggestion
            cmd.append('-DCMAKE_MAKE_PROGRAM:FILEPATH=' + ninja_executable_path)
```

henryiii:
The problem with the "2016" test (which has nothing to do with the fix in this PR, but only with the addition of a test that requires MSVC and not GCC from chocolaty) is that we start with the "ninja" MSVC generator, but do not enforce that specific MSVC is found.

henryiii:
Looks like we could use this: https://discourse.cmake.org/t/feature-request-give-me-a-way-to-specify-the-msvc-toolset-when-using-g-ninja/1332/8

This should probably not be mixed with this PR, though. I'll work around it for the moment.

henryiii:
Also, windows-2016 is about to be removed by GHA (in March, I think), so we should find a way to hit this problem without it, or solve it by then. ;)

PR #640, opened 2022-01-28
ci: drop unneeded job

henryiii:
The extra job shouldn't be needed anymore, and was missing a Python 3.5 initialize anyway (since it's past EOL it's not installed by default in GHA runners anymore), causing it to be broken.

We also don't need to run on 3.9 for win & macOS, we already do 2.7 (macOS), 3.5, and 3.10.


henryiii:
(I've fixed it without removing it in #634, but let's just remove it, since it's redundant now).

PR #641, opened 2022-01-28
chore: add pylint

henryiii:
(Re)enables PyLint, and fixes issues reported (or adding explicit skips).

Builds on #634.

- chore: add pylint and fix issues


jcfr:
Thanks for resuscitating this PR 🙏

Considering the amount of changes, I suggest to keep the transition to setuptools_scm, and the style fixes as two separate PRs.

henryiii:
Yes, #634 is just the setuptools_scm transition, and this is PyLint. I wanted pyproject.toml to configure PyLint, so I needed to build on that one. This will come out of draft when #634 is merged.

Issue #650, opened 2022-02-02
GCC is automatically detected on Windows instead of MSVC


planetmarshall:
This appears to be a recent change as some CI builds have started failing and the scikit-build version is the only thing that has changed.

Reproduced on Azure Devops windows-2019 host (agent version 2.198.2) Using cibuildwheel 2.3.0

## Log from failing build (scikit-build 0.13)

```
2022-02-02T20:00:32.1865951Z -- The C compiler identification is GNU 8.1.0
2022-02-02T20:00:32.5689077Z -- Detecting C compiler ABI info
2022-02-02T20:00:33.4125739Z -- Detecting C compiler ABI info - done
2022-02-02T20:00:33.4369218Z -- Check for working C compiler: C:/ProgramData/chocolatey/bin/gcc.exe - skipped
2022-02-02T20:00:33.4378665Z -- Detecting C compile features
2022-02-02T20:00:33.4392321Z -- Detecting C compile features - done
2022-02-02T20:00:38.2725206Z -- The CXX compiler identification is GNU 8.1.0
```

## Log from working build (scikit-build 0.12)

```
2022-02-02T17:42:38.5825398Z -- The C compiler identification is MSVC 19.29.30139.0
2022-02-02T17:42:38.6379084Z -- Detecting C compiler ABI info
2022-02-02T17:42:45.2473535Z -- Detecting C compiler ABI info - done
2022-02-02T17:42:45.2500227Z -- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/Tools/MSVC/14.29.30133/bin/Hostx86/x64/cl.exe - skipped
2022-02-02T17:42:45.2510064Z -- Detecting C compile features
2022-02-02T17:42:45.2522927Z -- Detecting C compile features - done
2022-02-02T17:42:49.2204422Z -- The CXX compiler identification is MSVC 19.29.30139.0
```

Placeholder until I can provide a minimal example.

## Workaround

Pin `scikit-build` at 0.12

henryiii:
The correct workaround is to activate the MSVC vars before building on windows, but okay. I thought this was specific to 2019+, since that was the version that added Ninja, and this is using ninja but finding GCC instead, and passing the check somehow. But it appears to be related to the addition of 2022 if pinning 0.12 "fixes" it.

henryiii:
[...]

PR #653, opened 2022-02-04
tests: include MSVC 2019

henryiii:
Investigating #650.


henryiii:
At least until windows-2016 is removed (Azure will need to be changed, too, that's only running on that image), we can reproduce #650. It looks like the generator discovery is working, but then it runs `-G Ninja`, and Ninja does _not_ find the MSVC image used for the discovery. Here's the command that ends up with GCC:

```
cmake 'D:\a\scikit-build\scikit-build\tests\samples\hello-pep518'
   -G Ninja
  '-DCMAKE_INSTALL_PREFIX:PATH=D:\a\scikit-build\scikit-build\tests\samples\hello-pep518\_skbuild\win-amd64-3.5\cmake-install\src\hello'
  '-DPYTHON_EXECUTABLE:FILEPATH=C:\Users\RUNNER~1\AppData\Local\Temp\build-env-1mw8i6_5\Scripts\python.exe'
  -DPYTHON_VERSION_STRING:STRING=3.5.4
  '-DPYTHON_INCLUDE_DIR:PATH=C:\hostedtoolcache\windows\Python\3.5.4\x64\Include'
  '-DPYTHON_LIBRARY:FILEPATH=C:\hostedtoolcache\windows\Python\3.5.4\x64\libs\python35.lib'
  -DSKBUILD:INTERNAL=TRUE
  '-DCMAKE_MODULE_PATH:PATH=C:\Users\RUNNER~1\AppData\Local\Temp\build-env-1mw8i6_5\lib\site-packages\skbuild\resources\cmake'
  '-DCMAKE_MAKE_PROGRAM:FILEPATH=C:\Users\RUNNER~1\AppData\Local\Temp\build-env-1mw8i6_5\lib\site-packages\ninja\data\bin\ninja'
  -DCMAKE_BUILD_TYPE:STRING=Release
```

henryiii:
Okay, I understand the bug, and can patch it, though longer term will likely need some further work. I also discovered several issues along the way that didn't help, and also will be fixed (I'll probably break the PR's out).

* GHA tests weren't setting the env vars on Windows.
* The check for generators wasn't checking to see what Ninja found - it always passed if Ninja passed, even if it found a different MSVC or GCC. I've made it fail if it finds something it's not looking for.


Now the actual regression: if `ninja` (python package) is installed, we are now forcing ninja to be the installed version. But we weren't forcing it for the check. The MSVC ninja finds the MSVC it is tied to, while the generic Ninja finds GCC. This leaves several possible solutions:

1. Don't force the Python package on Windows. This will restore the old behavior - but if you do end up using the Python package, it moves between invocations, so breaks. I _think_ you'll pretty much never use it, though, as many MSVC 2017's ship with it, and all future ones do.
2. Include the forcing in the check too. This should fall back on the next generator, the "native" one, which should work.


In any case, I think users should be prompted to place `"ninja; platform_system!='Windows'",` in their pyproject.toml's. There is generally no need to use the Python package since MSVC now ships with Ninja.

Long term, if we can improve the search to properly find MSVC from the generic Ninja, this won't be an issue.

I'll try Option 2 for 0.13.1, but if it doesn't work, I'm fine to go with option 1, which restores 0.12.0 behavior (with the above mentioned fixes).

skbuild/platform_specifics/abstract.py
```diff
@@ -190,7 +199,7 @@ def _generator_discovery_status_msg(_generator, suffix=""):
             outer = "-" * 80
             inner = ["-" * ((idx * 5) - 3) for idx in range(1, 8)]
             print(outer if suffix == "" else "\n".join(inner))
-            print("-- Trying \"{}\" generator{}".format(_generator.description, suffix))
+            print("-- Trying \"{}\" generator{} {}".format(_generator.description, suffix, _generator.args))
```
henryiii, line 202:
```suggestion
            print("-- Trying \"{}\" generator{}".format(_generator.description, suffix))
```

henryiii:
This fixes the problem by doing what the search was originally supposed to do - the Ninja and NMake generators fail if they can't find the MSVC that was requested. However, this is very slow - our 2016 tests jump from ~20 minutes or so to an hour to complete. And each run of Ninja/NMake is running exactly the same thing, it's just providing different failure conditions. I can instead modify the search to do Ninja (MSVC only), MSVC files (2022, 2019, etc), NMake, then maybe Ninja (anything). That would behave much more like the original 0.12 search and also fix this issue (by not using Ninja if it finds GCC). The only downside is that if ninja finds an older MSVC, that will be chosen, but that was true before, and should be quite rare (I hope).

henryiii:
I moved NMake to the end, which improves the job from 59 minutes to 55 minutes. Guess we'll have to be content with that for now.I can also split the job in two (3.5 on one, 3.10 on another) to further improve time to completion, but `windows-2016` is about to be removed, so probably not worth the effort. I really wish we could replicate the buggy behavior on one of the newer images! Mayne GCC isn't installed by default there?

henryiii:
We already find MSVC to activate it, so we probably could piggy-back on that for this. But this is fine for a fix to the immediate problem, something to consider for a future versions. Ideally after we drop Python 2.7-3.5 and MSVC<2015, which will really simplify this.

PR #656, opened 2022-02-07
docs: prepare for 0.13.1

henryiii:


henryiii:
`windows-2016` is in brown-out today.

PR #657, opened 2022-02-08
docs: starting changelog for 0.14

henryiii:


henryiii:
Was going to wait until 0.14 was a bit closer, but this fixes a linting error (and moving something from #477 that ended up in the wrong spot), so will merge sooner.

PR #673, opened 2022-03-05
fix library dedection for pypy 3.9

rkaminsk:
The library name in pypy 3.9 has changed. The '-c' moved to the end. This PR patches the candidate matrix to "make it work". It doesn't really fit nicely in the current matrix, though.

### DIFF
--- a/skbuild/cmaker.py
+++ b/skbuild/cmaker.py
@@ -482,9 +482,11 @@ class CMaker(object):
 
             candidate_lib_prefixes = ["", "lib"]
 
+            candidate_suffixes = [""]
             candidate_implementations = ["python"]
             if hasattr(sys, "pypy_version_info"):
                 candidate_implementations = ["pypy-c", "pypy3-c", "pypy"]
+                candidate_suffixes.append("-c")
 
             candidate_extensions = [".lib", ".so", ".a"]
             # On pypy + MacOS, the variable WITH_DYLD is not set. It would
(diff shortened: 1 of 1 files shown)


mattip:
Where does `sysconfig.get_config_var('LDLIBRARY')` fail to return the correct link library?

rkaminsk:
> Where does `sysconfig.get_config_var('LDLIBRARY')` fail to return the correct link library?

scikit-build is not using `LDLIBRARY`. But I also tried if it would work using it. Unfortunately, the value reported by the sysconfig module does not include the version number and I could not use it.

```
$ ./pypy3.9-v7.3.8-linux64/bin/pypy -m sysconfig | grep LDLIB
	LDLIBRARY = "libpypy3-c.so"

$ ls -1 ./pypy3.9-v7.3.8-linux64/bin/lib*
./pypy3.9-v7.3.8-linux64/bin/libpypy3.9-c.so
./pypy3.9-v7.3.8-linux64/bin/libpypy3.9-c.so.debug
```

mattip:
Ahh, right, that was a bug in the release and was fixed. It will be part of this week's release.

rkaminsk:
> Ahh, right, that was a bug in the release and was fixed. It will be part of this week's release.

Okay, then we should wait to merge this PR. I will have a look again once the new release is out. Even though the current state should also work with the upcoming release.

rkaminsk:
@mattip, I can confirm that LDLIBRARY is set correctly in the nightly build. But there is no updated release yet.

@henryiii, this pact tries to avoid guessing the library path by constructing it from config vars. If this fails, it falls back to the matrix to guess the path. The matrix has also been adjusted to work with newer pypy releases. Just in case. What do you think?

mattip:
> But there is no updated release yet.

I will be releasing a new version once CPython releases the upcoming [security fix release](https://discuss.python.org/t/can-a-fix-for-cve-2022-26488-be-available-sooner-than-april-4th/14175) (March 15?). I want to make sure I got all the fixes into PyPy.

henryiii:
Fantastic, thank you very much!

PR #675, opened 2022-03-07
Ensure that each package has a separate list of data elements.

vyasr:
Currently when scikit-build copies files from the source tree to the build tree (files that do not need to be compiled, e.g. pure Python files or Cython headers) it uses the `package_data` dict as a way to keep track of these files since data files are subject to the same requirements at this stage. `skbuild` assumes that modifying the `package_data` dict in place is safe [here](https://github.com/scikit-build/scikit-build/blob/master/skbuild%2Fsetuptools_wrap.py#L978) because it deep copies the `package_data` parameter passed to setup.py [here](https://github.com/scikit-build/scikit-build/blob/master/skbuild%2Fsetuptools_wrap.py#L513). However, deep copying is not sufficient in the scenario where the calling code constructs the input dict such that all values are actually references to the same dictionary, e.g.
```
files = ['foo', 'bar']
...
setup(..., package_data={package: files for package in packages},...)
```
or 
```
setup(..., package_data=dict.fromkeys(packages, ['foo', 'bar']), ...)
```
`setuptools.setup` appears to have no problem with these inputs, so `scikit-build` should support them as well. The safer approach here is to use a dict comprehension that explicitly copies each list. The lists will not be nested, so this approach is sufficient to guarantee that each list is unique and can be safely modified.

### DIFF
--- a/skbuild/setuptools_wrap.py
+++ b/skbuild/setuptools_wrap.py
@@ -510,7 +510,10 @@ def setup(*args, **kw):  # noqa: C901
 
     packages = kw.get("packages", [])
     package_dir = kw.get("package_dir", {})
-    package_data = copy.deepcopy(kw.get("package_data", {}))
+    package_data = {
+        k: v.copy()
+        for k, v in kw.get("package_data", {}).items()
+    }
 
     py_modules = kw.get("py_modules", [])
     new_py_modules = {py_module: False for py_module in py_modules}
--- a/skbuild/setuptools_wrap.py
+++ b/skbuild/setuptools_wrap.py
@@ -5,6 +5,7 @@ from distutils and setuptools.
 from __future__ import print_function
 
 import argparse
+import copy
 import json
 import os
 import os.path
@@ -509,7 +510,7 @@ def setup(*args, **kw):  # noqa: C901
 
     packages = kw.get("packages", [])
     package_dir = kw.get("package_dir", {})
-    package_data = {k: v.copy() for k, v in kw.get("package_data", {}).items()}
+    package_data = {k: copy.copy(v) for k, v in kw.get("package_data", {}).items()}
 
     py_modules = kw.get("py_modules", [])
     new_py_modules = {py_module: False for py_module in py_modules}
--- a/skbuild/setuptools_wrap.py
+++ b/skbuild/setuptools_wrap.py
@@ -5,7 +5,6 @@ from distutils and setuptools.
 from __future__ import print_function
 
 import argparse
-import copy
 import json
 import os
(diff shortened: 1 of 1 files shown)


jcfr:
For additional context:

Commit https://github.com/scikit-build/scikit-build/commit/c3a19bacd93898b929faee8c99966aa1503567de from 2018 added relevant tests and introduced the use of `deepcopy` along side with `_consolidate_package_data_files`:

```diff
-    package_data = kw.get('package_data', {}).copy()
+    package_data = copy.deepcopy(kw.get('package_data', {}))
```
_Source: https://github.com/scikit-build/scikit-build/commit/c3a19bacd93898b929faee8c99966aa1503567de#diff-31c811f03446fcfe70dbf03560cef0781f582d6f5d66334b76b2282e2e0b8733L386-R388_

For comparison, this pull request introduces the following changes:

```diff
-    package_data = copy.deepcopy(kw.get("package_data", {}))
+    package_data = {k: copy.copy(v) for k, v in kw.get("package_data", {}).items()}
```
_Source: https://github.com/scikit-build/scikit-build/pull/675/files#diff-31c811f03446fcfe70dbf03560cef0781f582d6f5d66334b76b2282e2e0b8733L513-R513_

vyasr:
To give some context about why these behavior are different:

My guess is that the first change you linked was made to fix an issue where if a user did
```
data_names = ['bar']
setup(..., package_data = {'foo': data_names}, ...)
```
and you just called `package_data.copy()` it would result in a shallow copy of the dict, so modifying `package_data['foo']` inside `skbuild` would lead to changes in the user-provided input list. Perhaps there were also other side effects within `skbuild` of such a shallow copy. My changes should still be safe in those cases, but they also address the more subtle case where a user does
```
data_names = ['bar']
setup(..., package_data = {'foo': data_names, 'baz': data_names}, ...)
```
In this scenario the `deepcopy` code will result in `skbuild` using a new object internally, so it will not persist changes to the caller's `data_names` variable. However, any changes to `package_data['foo']` inside `skbuild` will also propagate to `package_data['bar']` because the `deepcopy` will see that in the original dictionary both values are references to the same list and will therefore propagate that reference structure to the deep copy.

vyasr:
You could safely do a deepcopy. I think that the API explicitly requires the values to be lists of strings, so anything that _required_ a deep rather than a shallow copy for safety would violate the existing API, but there's probably no harm in future-proofing in case that changes.

jcfr:
Thanks @vyasr and @henryiii for following up on this :pray:

jcfr (approved):

PR #678, opened 2022-03-14
fix: do byte-compilation in install_lib

pekkarr:
`install_lib.run()` in [setuptools](../../../pypa/setuptools/blob/v60.9.3/setuptools/command/install_lib.py#L12) expects that the `outfiles` list is returned from `install_lib.install()`, otherwise byte-compilation is skipped.


henryiii:
Thanks!

Issue #698, opened 2022-04-12
Windows: error does result in exit 1?


tdegeus:
My build error on Windows does not lead to a CI failure : https://github.com/tdegeus/FrictionQPotSpringBlock/runs/5992549310?check_suite_focus=true

henryiii:
From a quick look, it seems the thing inside the SystemExit / sys.exit call is not an int like it should be.

tdegeus:
Indeed. I have the impression that there is a `try` `except` somewhere that prints something under `except` but that does not throw or exit in the end. The thing is, I cannot to easily estimate where that happens, as, because of the same reason there is not traceback (in general that is a bit of problem when `sys.exit` is used over `raise`).

henryiii:
You are replacing the bash command. The exit code is 1, but the following command `rm` succeeds, so the command passes. If you use the default shell, that errors immediately, but not if you replace the shell. You need to pass `-e` to bash.

tdegeus:
Thanks. So you say it would fail with 
```yaml
    defaults:
       run:
         shell: bash -e -l {0}
```
It just surprises me that I don't think I have had this problem before.

tdegeus:
I hear what you say, but now that I'm paying extra attention: in another repo I have 
```yaml
    - name: Run tests
      working-directory: tests
      run: |
        python foo.py
        python bar.py
```
In this case the CI fails when `foo.py` fails. `bar.py` is not even ran. True that happens on Ubuntu.

henryiii:
scikit-build (classic) after #1185 is now a thin wrapper around scikit-build-core's setuptools plugin. Closing all open issues; if this still applies, please reopen or open an issue on scikit-build-core. Thanks!

PR #699, opened 2022-04-13
fix: slightly better exits, more consistent

henryiii:
See #698.


henryiii:

PR #700, opened 2022-04-14
Fix distutils usage in Python 3.10

SuperSandro2000:


tests/__init__.py
```diff
@@ -1,6 +1,10 @@
 # -*- coding: utf-8 -*-
 
-import distutils
+try:
+    import distutils.dir_util
+    import distutils.sysconfig
+except ImportError:
+    import distutils  # Python < 3.10
```
henryiii, line 7:
```suggestion
import setuptools

import distutils.dir_util
import distutils.sysconfig
```

SuperSandro2000: I can't test if that works with older python versions.

henryiii: It should work with older Pythons, but likely not older setuptools. I think be need both the setuptools-first import, and the try/except.

SuperSandro2000: Yeah, that sounds more plausible but why do we need setuptools? CI and my local tests pass without it.

henryiii: Because you are always supposed to import setuptools before distutils. It patches distutils when you import it. Most builders (like pip) do this for you already, but it's good to always do it first in case they don't.

SuperSandro2000: amended and pushed

tests/__init__.py
```diff
@@ -1,6 +1,13 @@
 # -*- coding: utf-8 -*-
 
-import distutils
+import setuptools
```
henryiii, line 3:
```suggestion
import setuptools  # noqa: F401
```

PR #688, opened 2022-04-04
feat: Drop Python < 3.6 / old msvc

henryiii:
Needed to bring our CI back. Python 3.6+, MSVC 2017+.

PR #704, opened 2022-04-30
feat: Parse setup.cfg or equivalent to extract options that matter to skbuild

### LINKED ISSUE #591: Support setup.cfg package definition
Hi,

When building using scikit build, I seem to be required to use
- setup.py
`setup(packages=['mypackage'])`

Rather than being able to declare this in setup.cfg
- setup.cfg
```
[options]
packages = mypackage
```

Sort of related --- #405 --- if we solved that then my issue would likely be resolved too.

rdbisme:
This should provide the parsing of all config files supported by setuptools. It extracts only the ones that `skbuild` inspects at early stage in the `setup()` function.

Close #591


rdbisme:
PS: Does this need to work with Python 2.7?

henryiii:
Not after #688.

docs/conf.py
```diff
@@ -36,34 +36,29 @@
...
-#needs_sphinx = '1.0'
+# needs_sphinx = '1.0'
 
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
-extensions = [
-    'sphinxcontrib.moderncmakedomain',
-    'sphinx.ext.autodoc',
-    'sphinx.ext.viewcode',
-    'sphinx_issues'
```
jcfr, line 47: Nitpick: I usually keep such list spread across multiple lines and alphabetically sorted to allow easier review.

henryiii: Black shouldn’t have been run on conf.py, we had decided to leave it unformatted. (I’m fine to format it, but not part of this PR, and it needs to be in-excluded in pre-commit. And adding a trailing comma would keep it per-line). So this file should be reverted.

henryiii: PS: Also a fan of making everything where order doesn't matter alphabetically sorted. :)

skbuild/setuptools_wrap.py
```diff
@@ -358,6 +363,29 @@ def _load_cmake_spec():
...
+        "entry_points",
+        "scripts",
+    )
+
+    def __init__(self, files: Optional[Iterable[str]] = None):
+        self.dist = Distribution()
+        self.dist.parse_config_files(files)
+
+    def to_dict(self) -> dict[str, Any]:
+        return {k: v for k in self.relevant_options if (v := getattr(self.dist, k))}
```
henryiii, line 386:
```suggestion
        return {k:  getattr(self.dist, k) for k in self.relevant_options if getattr(self.dist, k, None)}
```

It does need to support Python 3.7 though. :D Also, this should not have worked, getattr without a default raises AttributeError, it does not return.

skbuild/setuptools_wrap.py
```diff
@@ -358,6 +363,29 @@ def _load_cmake_spec():
...
+        "data_files",
+        "entry_points",
+        "scripts",
+    )
+
+    def __init__(self, files: Optional[Iterable[str]] = None):
+        self.dist = Distribution()
+        self.dist.parse_config_files(files)
+
+    def to_dict(self) -> dict[str, Any]:
```
henryiii, line 385:
```suggestion
    def to_dict(self) -> Dict[str, Any]:
```

skbuild/setuptools_wrap.py
rdbisme: I think this works because `Distribution` initializes the attributes to `None` (i've been using the fork for a little while). Anyway the edit is better.

rdbisme:
> > I'm guessing this doesn't work with package = find:?
> 
> I think it should. I think [there's nothing specific in `setuptools.setup` that handles it](https://github.com/pypa/setuptools/blob/1e5e8df508d6304d40a5ebb7b87d37e12854317b/setuptools/__init__.py#L83-L87). So I suppose it's handled by `distutils`?

@henryiii you're right, this doesn't work for `find:`. Any idea where that is handled in the `setuptools` machinery such that I can give a look?

Also, can that be done in a future MR?

henryiii:
The ideal way to do this I think is to use `run_setup`, `https://github.com/python/cpython/blob/2efe18bf277dd0f38a1d248ae6bdd30947c26880/Lib/distutils/core.py#L170`, which would leave the parsing to setuptools and then we could query the options from there. The method in this PR also won't support pyproject.toml configuration, which I'd like to support as well if possible. @abravalheri might have some comment on the best way to pull configuration before running the build?

abravalheri:
@henryiii is right, using something like:

```python
from setuptools.dist import Distribution  # <-- setuptools needs to be imported first
from distutils.core import run_setup

if os.path.exists(setup_script):
   dist = run_setup(setup_script, stop_after="init")
else:
   dist = Distribution({"script_name": "%notset%", "script_args": []})

dist.parse_config_files()
# dist attributes has now be filled in by config files.
```

Should do the trick... However note that people using `SETUPTOOLS_USE_DISTUTILS=stdlib` can experience some issue depending on how they write `setup.py` scripts, because the original version of distutils has worse encoding handling.

rdbisme:
What I'd like to achieve is the same behaviour as a `pep517` compliant frontend. I.e. running something like a `resolve_config()` that automatically understands which config file to check based on `pyproject.toml`, doing the necessary runtime tasks to populate the `Distribution` object with the metadata we need to inspect. 

The solution @abravalheri you suggest requires coding the logic to identify the `setup_script`, doesn't it? 

Maybe I'll check again how `build` package works and see if I came up with a good idea.

abravalheri:
> Maybe I'll check again how build package works and see if I came up with a good idea.

You are probably looking for [pep517.meta.load](https://github.com/pypa/pep517/blob/c45334a7bc44722d5f9ef0d658b5307747ed04e8/pep517/meta.py#L63) (I don't know how public this function is considered to be).

The caveat is that this function does run part of the build process, specifically [prepare_metadata_for_build_wheel](https://peps.python.org/pep-0517/#prepare-metadata-for-build-wheel). It is very heavy weight also...

> The solution @abravalheri you suggest requires coding the logic to identify the setup_script, doesn't it?

You can assume `setup.py`... The important part is to check if a `setup.py` script actually exists at the root of the project or not.

henryiii:

PR #712, opened 2022-05-17
Support FindPython.

### LINKED ISSUE #506: FindPython support
I am currently working on adding full CMake 3.12+'s FindPython support to pybind11. Would it be possible to support for FindPython in Scikit-Build's extensions, too?

hameerabbasi:
Closes #506 and closes #516 and closes #553.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,7 +9,7 @@ Scikit-build 0.15.0
 ===================
 
 This release is the final (again) release for Python < 3.6 and MSVC<2017. Support
-for FindPython from CMake 3.12+ was added, inluding FindPython2.
+for FindPython from CMake 3.12+ was added, including FindPython2.
 
 New Features
 ------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,20 @@ Release Notes
 This is the list of changes to scikit-build between each release. For full
 details, see the commit logs at http://github.com/scikit-build/scikit-build
 
+Scikit-build 0.15.0
+===================
+
+This release is the final (again) release for Python < 3.6 and MSVC<2017. Support
+for FindPython from CMake 3.12+ was added, inluding FindPython2.
+
+New Features
+------------
+
+* Add support for FindPython (including 2 and 3).
+
+
+Thanks :user:`hameerabbasi` for the contribution. See :pr:`712`.
+
 Scikit-build 0.14.1
 ===================
 
--- a/tests/samples/hello-numpy/CMakeLists.txt
+++ b/tests/samples/hello-numpy/CMakeLists.txt
@@ -5,7 +5,7 @@ project(hello-pybind11 VERSION "0.1")
 # Define CMAKE_INSTALL_xxx: LIBDIR, INCLUDEDIR
 include(GNUInstallDirs)
 
-find_package(Python3 REQUIRED COMPONENTS Development NumPy)
+find_package(Python REQUIRED COMPONENTS Development NumPy)
 
 # Fetch pybind11
 include(FetchContent)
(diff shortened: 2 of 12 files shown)


tests/samples/hello-numpy/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+include(GNUInstallDirs)
+
+find_package(Python REQUIRED COMPONENTS Development NumPy)
+
+# Fetch pybind11
+include(FetchContent)
+
+FetchContent_Declare(
+  pybind11
+  URL https://github.com/pybind/pybind11/archive/refs/tags/v2.6.2.tar.gz
```
henryiii, line 15: Guessing this (and the ...3.18 above) were copied from another example? We really should be update the pins/limits. Doesn't have to be in this PR, though, just noting for later.

tests/samples/hello-numpy/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+         pybind11::module
+)
+
+target_compile_definitions(${python_module_name}
+                           PRIVATE VERSION_INFO=${EXAMPLE_VERSION_INFO})
+
+install(TARGETS ${python_module_name} DESTINATION .)
+
+# Quiet a warning, since this project is only valid with SKBUILD
+set(ignoreMe "${SKBUILD}")
```
henryiii, line 37: Technically, we now have a _lot_ of variables that are going to produce warnings (this is why I haven't applied this sort of a fix yet. But I think it's fine, and I have a long term idea for fixing these warnings). Not sure it makes sense to just quiet this one anymore. But again, not for this PR.

tests/test_numpy.py
```diff
@@ -0,0 +1,16 @@
...
+BASE = os.path.dirname(DIR)
+
+
+def test_pep518(pep518):
+    subprocess.check_call([sys.executable, "-m", "build", "--wheel"], cwd=HELLO_NUMPY)
+
+
+def test_dual_pep518(pep518):
+    subprocess.check_call([sys.executable, "-m", "build", "--wheel"], cwd=HELLO_NUMPY)
+    subprocess.check_call([sys.executable, "-m", "build", "--wheel"], cwd=HELLO_NUMPY)
```
henryiii, line 16:
```suggestion
```

I don't think we need this test in this case, just making sure it works is fine (and these tests are expensive).

tests/test_numpy.py
```diff
@@ -0,0 +1,16 @@
+import os
+import subprocess
+import sys
+
+DIR = os.path.dirname(os.path.abspath(__file__))
+HELLO_NUMPY = os.path.join(DIR, "samples/hello-numpy")
+BASE = os.path.dirname(DIR)
+
+
+def test_pep518(pep518):
```
henryiii, line 10:
```suggestion
def test_pep518_findpython(pep518):
```

hameerabbasi:
I've gone through and applied the simple fixes suggested in the review.

henryiii (approved):
I think this is good. We sadly will produce unused variable warnings, but that's unavoidable for now, and I think I have a general solution to be applied later. I think we should get this in so we get it into a final Python 2 release. Thanks!

CHANGES.rst
```diff
@@ -5,6 +5,20 @@ Release Notes
 This is the list of changes to scikit-build between each release. For full
 details, see the commit logs at http://github.com/scikit-build/scikit-build
 
+Scikit-build 0.15.0
+===================
+
+This release is the final (again) release for Python < 3.6 and MSVC<2017. Support
+for FindPython from CMake 3.12+ was added, inluding FindPython2.
```
henryiii, line 12:
```suggestion
for FindPython from CMake 3.12+ was added, including FindPython2.
```

henryiii:
We'll need to rebase #485 on this, so let's get it in. Thank you!

Issue #358, opened 2018-09-05
py_modules with package_dir


bgermann:
py_modules with package_dir does not work as expected. On a bdist such modules are not included as with setuptools. Example: https://github.com/silnrsi/graphite/blob/97bbe445f/setup.py

jcfr:
Thanks for reporting the issue. We should definitely fix this.

Could you create a test (and companion sample project with relevant) to reproduce the problem ?

You could for example look at: https://github.com/scikit-build/scikit-build/tree/master/tests

Note: I will be traveling until Sept 29 with limited access to email

henryiii:
scikit-build (classic) after #1185 is now a thin wrapper around scikit-build-core's setuptools plugin. Closing all open issues; if this still applies, please reopen or open an issue on scikit-build-core. Thanks!

PR #713, opened 2022-05-17
Update setuptools_wrap.py

aws-taylor:
Candidate fix for https://github.com/scikit-build/scikit-build/issues/358 - The issue is that 'package_dir' in setup.py may result in 'package' being substantially different from the directory prefix of 'src_module_file'. To fix this, instead of using all of 'src_module_file', instead just retain the base .py file, but also prefix with the expected package path.

### DIFF
--- a/skbuild/setuptools_wrap.py
+++ b/skbuild/setuptools_wrap.py
@@ -13,6 +13,7 @@ import platform
 import stat
 import sys
 import warnings
+import functools
 from contextlib import contextmanager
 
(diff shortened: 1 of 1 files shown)


aws-taylor:
Drat, this approach isn't going to work. It seems like the code above correctly copies the files in the case of a `build`, but under `bdist_wheel` is seems some other path is being used to find files - possibly related to how the SOURCES manifest is created, and the above change subverts that.

PR #716, opened 2022-05-18
tests: enable cygwin skipped tests

henryiii:


henryiii:
Looks like CMake doesn't build under Cygwin. Likely will need to be skipped in pyproject.toml's for now.

DWesl:
Locally, it appears the python `cmake` package defines `_WIN32` or `__WIN32__` on Cygwin, which breaks its bundled curl build.  Is there some way to tell it to use the system curl (`libcurl-devel`)?

henryiii:

PR #728, opened 2022-07-01
chore: enable basic mypy

henryiii:
This was sitting around, let's go ahead and see what the diff looks like. I was playing with monkeytype.

- chore: update mypy settings a bit
- chore: run monkeytype, update old workarounds

Looks like monkey type was too specific, some need generalization. :)

Edit: Worked on the added types, and got mypy running by default.

A few notes:

* There are several type ignores due to https://github.com/python/mypy/issues/12344
* I didn't work out all the details of Distribution, so there are lots of type ignores there
* There are a few minor cleanups noticed by mypy, like random trailing commas

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,9 +41,8 @@ module = [
   "wheel.*",
   "setuptools.logging",
   "setuptools.command.*",
-  "ninja",
-  "cmake",
   "numpy",
+  "distro",
 ]
 ignore_missing_imports = true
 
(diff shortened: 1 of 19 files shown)


pyproject.toml
```diff
@@ -26,15 +26,38 @@ profile = "black"
...
+warn_redundant_casts = true
+warn_unused_ignores = true
+no_implicit_reexport = true
+disallow_any_generics = true
+disallow_incomplete_defs = true
+
+# disallow_subclassing_any = true 4
+# warn_return_any = true 5
+# disallow_untyped_calls = true 39
+# disallow_untyped_defs = true 61
```
henryiii, line 50: These are the remaining failures that need addressing before this is fully strictly typed. Started a bit on `disallow_untyped_defs` but it looks like the comments lie a bit, so will take a bit of work. Best to get this in as a strict improvement, I think, then work on these separately.

PR #731, opened 2022-07-15
replace guide for setup_requires with in-tree backend

maxbachmann:
setup_requires is deprecated in setuptools and it is possible to achieve the same using an `in-tree backend`.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -381,6 +381,7 @@ For this purpose place the following configuration in your ``pyproject.toml``::
     [build-system]
     requires = [
       "setuptools>=42",
+      "packaging",
       "scikit-build",
       "ninja; platform_system!='Windows'"
     ]
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -395,23 +395,20 @@ then you can implement a thin wrapper around ``build_meta`` in the ``_custom_bui
     prepare_metadata_for_build_wheel = _orig.prepare_metadata_for_build_wheel
     build_wheel = _orig.build_wheel
     build_sdist = _orig.build_sdist
+    get_requires_for_build_sdist = _orig.get_requires_for_build_sdist
 
-    def _get_requires():
+    def get_requires_for_build_wheel(self, config_settings=None):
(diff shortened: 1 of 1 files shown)


maxbachmann:
todo the usage of LegacyVersion needs to be replaced, since it was removed from packaging. I think something like `get_cmake_version_tuple` would be useful.

maxbachmann:
I am unsure when `get_requires_for_build_sdist` is called. Is this called to collect requirements to convert from source to sdist, or to build from the sdist?

henryiii:
Yes, that's right. `get_requires_for_build_sdist` is called when making SDists.

henryiii:
And this is a really interesting idea for scikit-build-core too...

maxbachmann:
In RapidFuzz I do the same for ninja/make. In addition I only install them when wheels are available, since the source build requires cmake to be available anyways. When installation fails due to compiler errors / missing cmake I provide a pure python fallback implementation: https://github.com/maxbachmann/RapidFuzz/blob/main/_custom_build/backend.py

henryiii:
Thanks! This technically does't increase our API footprint (since the use of the these parts of the API were already there), so I think it's fine to merge, though I'm hoping scikit-build's own backend will do this soon.

maxbachmann:
> I'm hoping scikit-build's own backend will do this soon.

An implementation inside scikit-build would likely always install cmake when it is not available yet. Note that this would break my usage of scikit-build. I do provide a pure Python fallback in case the compilation breaks. However this breaks when the build has a non pure Python dependency, which fails to install. Right now I hack this in the following way:
```python
if cmake_available:
    use system version
elif cmake_wheel_available:
    install wheel
else:
    use pure Python version
```
It would be great to add the cmake/ninja dependencies as a install extra `scikit-build[cmake]`.

henryiii:
No, what I mean is that `build-backend = "scikit_build_core.build"` will do pretty much exactly the recommendation in this PR by default, but would be customizable via `[tool.scikit-build.cmake]` settings.

Issue #697, opened 2022-04-12
Error when installing in development mode


romankempt:
**Describe the bug**
Installing a project via `pip install .` works, but fails when trying to install in development mode (either `pip install -e .` or `python setup.py develop`).

The error is triggered here:
```
Traceback (most recent call last):
  File "/mnt/c/Users/rkempt/Repositories/hetbuilder/setup.py", line 39, in <module>
    setup(
  File "/home/rkempt/miniconda3/envs/hetbuilder/lib/python3.9/site-packages/skbuild/setuptools_wrap.py", line 701, in setup
    package_file = os.path.join(package_dir[package], package_file)
KeyError: ''
```

The source code of the relevant function is here:

```python
    if developer_mode:
        # Copy packages
        for package, package_file_list in package_data.items():
            for package_file in package_file_list:
                package_file = os.path.join(package_dir[package], package_file)
                cmake_file = os.path.join(CMAKE_INSTALL_DIR(), package_file)
                if os.path.exists(cmake_file):
                    _copy_file(cmake_file, package_file, hide_listing)
```

I suspect some package variable is not set correctly for development mode?

**To Reproduce**

My `setup.py` looks like this:

```python
setup(
    name="hetbuilder",
    version=version,
    author="Roman Kempt",
    author_email="roman.kempt@tu-dresden.de",
    description="A tool to build heterostructure interfaces based on coincidence lattice theory.",
    long_description=long_description,
    license="MIT",
    url="https://github.com/romankempt/hetbuilder.git",
    download_url="https://github.com/romankempt/hetbuilder.git",
    packages=find_packages(),
    package_data={"": ["*.xyz", "CMakeLists.txt"]},
    # package_dir={"": ""},
    cmake_install_dir="hetbuilder",
    include_package_data=True,
    scripts=["bin/hetbuilder"],
    install_requires=[
        "numpy",
        "scipy",
        "spglib",
        "matplotlib",
        "ase",
        "networkx",
        "pretty_errors",
        "rich",
        "typer",
        "pybind11",
    ],
    classifiers=[
        "Operating System :: OS Independent",
        "Programming Language :: Python :: 3.9",
        "Programming Language :: C++",
        "Topic :: Scientific/Engineering :: Chemistry",
        "Topic :: Scientific/Engineering :: Physics",
    ],
    cmake_args=[
        "-DCMAKE_BUILD_TYPE={}".format("RELEASE"),  # not used on MSVC, but no harm,
    ],
[...]

Issue #436, opened 2019-07-07
Document relationship between library install destination and python package


jcfr:
See https://github.com/scikit-build/scikit-build/issues/434#issue-460839996

kratsg:
This bit me a bit, and I'm struggling to understand how this sort of interacts with each other (especially in editable installs). It would be great to have clearer documentation here :) ./cc @henryiii as I pinged him offline about this and it was not obvious why this was not behaving as expected.

henryiii:
scikit-build (classic) after #1185 is now a thin wrapper around scikit-build-core's setuptools plugin. Closing all open issues; if this still applies, please reopen or open an issue on scikit-build-core. Thanks!

Issue #370, opened 2018-12-05
Add support for "editable install" to the "Advanced Usage" section


jcfr:
See https://setuptools.readthedocs.io/en/latest/setuptools.html#develop-deploy-the-project-source-in-development-mode or https://pip.pypa.io/en/stable/reference/pip_install/#editable-installs

Suggested by @Erotemic

alexpearce:
Could I ask for some clarification on editable installs here?

Using `pip install -e .`, I see the files installed by CMake are placed in the `_skbuild/{platform}/cmake-install` folder. The same paths are listed in the generated `{package}.egg-info/SOURCES.txt` file. However I need the paths to the generated files at runtime, and either guessing the `_skbuild` path or parsing the `SOURCES.txt` file feels too ad-hoc.

The impression I get reading around the issues here is that the files inside `cmake-install` should be copied somewhere more 'obviously accessible', but I don't see that happening. Is that indeed expected?

I'm not specifying any `package_data`-related kwargs in my `setup` call (maybe I need to?), and in case it's a bug all of the files installed by CMake are in subfolders (`include/` and `lib/`).

alexpearce:
For future readers: my comment above is similar #434 (and #436 by association). I was installing things from my `CMakeLists.txt` to `lib`, but if I instead install them to `${PROJECT_NAME}/lib` then `pip install -e .` places the files in the package directory. For example my cmake config now looks like

```cmake
project(my_package)

add_library(${PROJECT_NAME} SHARED something.cxx)

install(TARGETS ${PROJECT_NAME} LIBRARY
        DESTINATION ${PROJECT_NAME}/lib
```

and `my_package` is the name of the package defined in `setup.py`, e.g.

```python
setup(
    name='my_package',
    # ...
)

jcfr:
Hi Alex,

PS: I can provide more details when I get access to a computer.

In the mean time, consider changing

DESTINATION ${PROJECT_NAME}/lib


Into


DESTINATION ${PROJECT_NAME}/my_package


Hth

Jc

alexpearce:
Your suggestion would install into `my_package/my_package/mylib.so`, right? (I defined `project(my_package)`, so `${PROJECT_NAME}` is equal to the name of my Python package.)

Is there a reason to prefer that over `my_package/mylib.so`?

ndevenish:
[...]

PR #737, opened 2022-08-07
Finer exceptions and hopes for editable installs.

claudioperez:
There does not seem to be a robust solution for editable installs (#370, #436, #546, #579, #697). The closest I've been able to get raises from `_classify_installed_files`, even though everything looks exactly as one might expect for an editable install. The only thing added here is a new exception class that gets raised in this situation so that the exception can be safely singled out and permitted, without catching every `SKBuildError`. Its not an ideal solution to the editable installation problem, but having more detailed exceptions sounds like an all-around plus.

### DIFF
--- a/skbuild/exceptions.py
+++ b/skbuild/exceptions.py
@@ -8,9 +8,10 @@ class SKBuildError(RuntimeError):
     project.
     """
 
+
 class SKBuildInvalidFileInstallationError(SKBuildError):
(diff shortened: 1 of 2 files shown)


henryiii:
Thanks! I think setuptools 64 broke whatever bit of editable install support we did have, FYI.

claudioperez:
No, thank you! This is much needed work you're doing here.

PR #739, opened 2022-08-15
fix: ensure absolute paths

henryiii:
Testing something.


henryiii:
Pretty sure the issue a user was having was not this, but rather that they were trying to access the static lib from manylinux (which is disallowed).

PR #743, opened 2022-09-03
fix(Cython): support FindPython

henryiii:
The language level of Cython depends on Cython version, not the host Python version. Avoiding this avoids requiring FindPythonLibs, which should stop breaking this when using FindPython.

Generalizing FindCython to use an existing discovery of Python and not trigger it's own.

CC @maxbachmann.


henryiii:
Used in https://github.com/maxbachmann/RapidFuzz/pull/256. We should be able to avoid the `Python_FIND_IMPLEMENTATIONS` setting as well in Scikit-Build.

PR #744, opened 2022-09-03
fix(pypy): include find impl setting

henryiii:
See #743. Makes this a tiny bit easier to use for PyPy.


henryiii:
I'd rather like to backport a design from scikit-build-core that fixes the unused variable warnings, then this will be nicer.

henryiii:
Might backport the warning fix (though for some variables it might not be ideal, since it won't override cached values, and some of them might change (virtualenv related locations, for example). But we should probably solve those separately anyway.

PR #748, opened 2022-09-09
Use stdlib ContextDecorator and os.makedirs

Skylion007:
These are some code shims that were support Python 2. Since it seems like scikit-build no longer supports them, might as well replace them with their Python Stdlib equivalents.


henryiii (approved):
Thanks! I think I started to do this, noticed the test that used kwargs, and put it off assuming it would be harder to fix.

PR #754, opened 2022-09-24
chore: simply strings a bit

henryiii:
A bit of cleanup, and this might be a bit more likely to work - the MODULE path wasn't absolute before. Investigating a missing module failure.


henryiii:
Wow, it does seem this breaks Windows...

henryiii:
Let's get this in as a tiny cleanup, then the diff should be simpler to see why Windows cares.

PR #769, opened 2022-10-08
fix: add mypy strict-mode checking

henryiii:
Simpler imports based on what's actually there.

### DIFF
--- a/skbuild/command/bdist_wheel.py
+++ b/skbuild/command/bdist_wheel.py
@@ -1,63 +1,37 @@
 """This module defines custom implementation of ``bdist_wheel`` setuptools
 command."""
 
-try:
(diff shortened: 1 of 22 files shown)


skbuild/command/bdist_wheel.py
```diff
@@ -1,65 +1,39 @@
 """This module defines custom implementation of ``bdist_wheel`` setuptools
 command."""
 
-try:
-    from wheel.wheelfile import WheelFile
-
-    _USE_WHEELFILE = True
-except ImportError:
-    from wheel import archive as _wheel_archive  # Not available with wheel >= 0.32.0
```
henryiii, line 9: No longer needed, we require wheel 0.32+ and have for a bit.

PR #797, opened 2022-11-18
Address Issue #676 

### LINKED ISSUE #676: Installations contain __init__.py in subdirectories of packages even if they are not explicitly requested
Given a package layout like so
```
    .
    ├── package
    |   ├── __init__.py
    │   ├── subpackage
    |       ├── __init__.py
    |       ├── subpackage_module.py
```
if I do not add a cython target or a library for subpackage, `subpackage/subpackage_module.py` is not copied into the install tree, but `subpackage/__init__.py` is. I would guess that this is because of some special treatment of `__init__.py` files.

aaron-bray:
Address #676.


henryiii:

PR #823, opened 2022-12-15
docs: Replace mailing list with GitHub Discussions board

jcfr:
* Simplify release guide by (1) referencing the existing announcement category available on the Discussions board and (2) removing the hard-coded template.

* Update `author_email` specified in `setup.py` removing google group email and instead listing emails for @jcfr, @henryiii and @thewtex.


setup.py
```diff
@@ -26,14 +26,14 @@
     long_description_content_type="text/x-rst; charset=UTF-8",
     long_description=readme + "\n\n" + history,
     author="The scikit-build team",
-    author_email="scikit-build@googlegroups.com",
+    author_email="jcfr@kitware.com;henryfs@princeton.edu;matt.mccormick@kitware.com",
```
henryiii, line 29: Is this actually supported? I know how to do it in PEP 621 config, not sure if this is right here though.

jcfr: Good catch, I add reviewed spec associated with `mailto:` URI scheme and assumed this would work.

After reviewing the documentation[^1], using `,` instead of `;` should be supported.

It turns out that by specifying the following:

```python
setup(
...
    author_email="jcfr@kitware.com, henryfs@princeton.edu, matt.mccormick@kitware.com",
...
)
```

it leads to a `METADATA` file with

```
Author-email: jcfr@kitware.com, henryfs@princeton.edu, matt.mccormick@kitware.com
```

but the HTML generated on PyPI doesn't include the other emails:

image

Source: https://test.pypi.org/project/jcfr-pypi-sandbox/






[^1]: https://packaging.python.org/en/latest/specifications/core-metadata/#author-email

jcfr: Now since the author email field is optional and specifying multiple email is not supported, I will remove it.

jcfr:
To follow-up, after removing support for python 3.6, we should be able to adopt backend like [hatchling](https://hatch.pypa.io/latest/build/#packaging-ecosystem) (requiring python >=3.7[^1]) allowing to specify multiple authors.

In the meantime, we will remove `author_email` from `setup.py`

[^1]: https://github.com/pypa/hatch/blob/hatchling-v1.11.1/pyproject.toml#L10

PR #824, opened 2022-12-15
fix: Windows ARM cross-compile

henryiii:
Should help get cross-compiles working.


henryiii:
I was able to daisy chain together a set of adaptors and cables (including VGA) to connect my Windows ARM box to a monitor. I was able to run Ninja (after reinstalling the ARM version of Python - the Microsoft Store doesn't seem to give the ARM version, but you can get it from Python.org). I'd say this is working.

thewtex:
Awesome!!

PR #831, opened 2022-12-20
fix function signature of get_requires_for_build_wheel

maxbachmann:
This fixes the function signature of `get_requires_for_build_wheel` to be compliant with pep517 (see https://github.com/pypa/setuptools/pull/3730)


henryiii:
Ahh, in the docs. Was confused at first! 😆

Issue #791, opened 2022-10-30
19 tests fail


yurivict:
Please see the log: [scikit-build-test-0.16.1.log](https://github.com/scikit-build/scikit-build/files/9896692/scikit-build-test-0.16.1.log)

henryiii:
Are these new failures compared to 0.15.0?

henryiii:
I do see the "exclude data" tests showing up. That means you have setuptools_scm installed while testing. You need setuptools_scm to build scikit-build, but it should not leak into the runtime requirements. Users should not be forced to install setuptools_scm when they install scikit-build, as setuptools_scm changes the way files are included by setuptools simply by being present.

yurivict:
When ```setuptools_scm``` has been removed now 13 tests fail.

> Are these new failures compared to 0.15.0?

0.15.0 also had 13 failed tests.

henryiii:
Excellent, as long as we aren't regressing. I'd like to work on the failures, but first goal is to make sure they don't grow. :) Do you know why there's no failed test summary at the end (`-ra` is specified in the pyproject.toml file)?

henryiii:
Here's my grep-based summary (of the 17 tests):

```
tests/test_broken_project.py::test_invalid_generator FAILED              [  2%]
tests/test_hello_cpp.py::test_hello_develop FAILED                       [ 22%]
tests/test_include_exclude_data.py::test_hello_sdist FAILED              [ 26%]
tests/test_include_exclude_data.py::test_hello_sdist_with_base FAILED    [ 27%]
tests/test_issue274_support_default_package_dir.py::test_install_command FAILED [ 27%]
tests/test_issue274_support_default_package_dir.py::test_test_command FAILED [ 28%]
tests/test_issue274_support_one_package_without_package_dir.py::test_install_command FAILED [ 28%]
tests/test_issue274_support_one_package_without_package_dir.py::test_test_command FAILED [ 28%]
tests/test_issue334_configure_cmakelists_non_cp1252_encoding.py::test_install_command FAILED [ 29%]
tests/test_issue401_sdist_with_symlinks.py::test_sdist_with_symlinks FAILED [ 34%]
tests/test_manifest_in.py::test_manifest_in_sdist FAILED                 [ 35%]
tests/test_numpy.py::test_pep518_findpython FAILED                       [ 36%]
tests/test_pep518.py::test_pep518 FAILED                                 [ 38%]
[...]

PR #837, opened 2023-01-10
tests: add markers for isolated & nosetuptoolsscm

henryiii:
- tests: add marker for nosetuptoolsscm
- tests: add isolated marker

See #791.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -94,4 +94,5 @@ markers = [
   "fortran: Fortran testing",
   "deprecated: These tests deprecated setuptools features",
   "nosetuptoolsscm: Requires setuptools_scm to not be installed",
+  "isolated: Downloads dependencies and sets up isolated environments",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -93,4 +93,5 @@ log_cli_level = "info"
 markers = [
   "fortran: Fortran testing",
   "deprecated: These tests deprecated setuptools features",
+  "nosetuptoolsscm: Requires setuptools_scm to not be installed",
 ]
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,3 +1,4 @@
+import importlib.util
 import os
 import subprocess
 import sys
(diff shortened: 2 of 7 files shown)


hroncok:
What if the nosetuptoolsscm tests were skipped when setuptools_scm is installed?

hroncok:
If I understand the docs correctly https://docs.pytest.org/en/7.1.x/example/markers.html

If we add something like this to conftest.py:

```py
try:
    import setuptools_scm
except ImportError:
    HAS_SETUPTOOLS_SCM = False
else:
    HAS_SETUPTOOLS_SCM = True
    del setuptools_scm
    del sys.modules["setuptools_scm"]

def pytest_runtest_setup(item):
    for marker in item.iter_markers(name="nosetuptoolsscm"):
        if HAS_SETUPTOOLS_SCM:
            pytest.skip("This test does not work when setuptools_scm is installed")
```

henryiii:
That's about how you'd do it (it's basically an inverted importorskip!), but to me, it seems that verifying that setuptools_scm is not present is normally helpful - you don't want a downstream package using scikit-build to be forced to use setuptools_scm, which unconditionally changes their build. This has helped at least one packager ensure that their build requirements were not leaked as install requirements, IIRC.

Of course, it should probably be a better failure than just some seemingly random tests failing.

henryiii:
Okay, I've improved the error message when the tests aren't skipped. It still runs till it hits one, but then it errors out immediately with a better error.

mgorny:
As a downstream packager, I'm completely confused by this. Does that mean that we're supposed to unmerge `setuptools_scm` before running the test suite (which — besides being very inconvenient for Gentoo users — is simply impossible because it's required by `hatch-vcs`)? Or are we supposed to skip these tests?

Also, `pytest.exit()` is horrible UX. It looks like the test suite suddenly terminated in the middle of test suite for no apparent reason, and the error is barely noticeable. I mean, if you don't look at exit status, at a first glance it looks as if the test suite succeeded.

PR #838, opened 2023-01-11
chore: improve extras granularity

henryiii:
Should help with downstream packaging, such as removing the need for https://copr-dist-git.fedorainfracloud.org/cgit/thrnciar/python-scikit-build/python-scikit-build.git/tree/Remove-test-deps-missing-in-Fedora.patch from https://github.com/scikit-build/scikit-build/issues/834.


docs/installation.rst
```diff
@@ -34,16 +34,8 @@ Dependencies
...
-
-.. include:: ../requirements.txt
-   :literal:
-
-The build time dependencies (also required for development) are:
-
-.. include:: ../requirements-dev.txt
-   :literal:
+The project has a few common Python package dependencies. These can be seen in
+``setup.py``.
```
henryiii, line 38:
```suggestion
The project has a few common Python package dependencies. These can be seen in
``setup.py`` and ``pyproject.toml``.
```

The old list is incorrect, things like `flake8` are not a requirement, and coverage was not a build time requirement. Nowadays, listing requirements in docs is not nearly as common.

setup.py
```diff
@@ -11,14 +11,36 @@
...
+        "sphinx>=4",
+        "sphinx-issues",
+        "sphinx-rtd-theme>=1.0",
+        "sphinxcontrib-moderncmakedomain>=3.19",
+    ],
+}
 
-with open("requirements-docs.txt") as fp:
-    doc_requirements = list(filter(bool, (line.strip() for line in fp)))
+extras_require["dev"] = extras_require["test"] + extras_require["cov"] + extras_require["doctest"]
```
henryiii, line 43: I'm leaving the dev requirement for now, though static dep lists make it harder to do these sort of combined requirements, so I'm focusing on listing the separate ones everywhere.

henryiii: Oh, actually, this was called "tests" or "requirements-dev" before, so maybe it would be best to just not add it at all, since we might want to remove it when moving to static configuration later.

henryiii:
```suggestion
```

PR #871, opened 2023-02-26
Stop assuming that pyx files are in the same directory as CMakeLists.txt

### LINKED ISSUE #707: UseCython.cmake enforces a directory layout on Cython projects
[This line](https://github.com/scikit-build/scikit-build/blob/0.15.0/skbuild/resources/cmake/UseCython.cmake#L213) in UseCython.cmake assumes that a CMakeLists.txt file exists alongside every pyx file. As a result, any Cython build managed with scikit-build must be structured such that each subpackage has its own CMakeLists.txt file that is then added via `add_subdirectory` from the parent directory's CMakeLists.txt file. This line should instead default to the current executed CMakeLists.txt directory (which could still be a subdirectory of the root) so that a project could use e.g. a single CMakeLists.txt to configure the whole project if desired.

vyasr:
This change allows users to leverage a single lists file for a project rather than needing to include a separate lists file in each subdirectory. Closes #707


vyasr:
I ran into a rather odd error with the test suite. It looks like `test_broken_project.py::test_hello_with_compileerror_fails` leaves the test environment in a broken state w.r.t. a nonfunctional `hello` package. As a result, my tests were initially failing because they generated a package with the same name. I could reproduce this locally by running `pytest -x -k "test_hello_with_compileerror_fails or 707"`, whereas `pytest -x -k "707"` passed. I fixed this in https://github.com/scikit-build/scikit-build/pull/871/commits/33af29f6225e759320c8893125a6ff56729ddb20 by renaming my package, but it might be good to clean up the broken package in that test if it isn't too difficult.

vyasr:
I think I'm misunderstanding how these tests are designed to be set up, or perhaps they're simply not designed to generate importable packages. I realized that using the installation locally was simply triggering an install into my host environment, not a virtual environment isolated to the test, which explains why my local results weren't matching CI. I have temporarily changed the test to use `build_ext --inplace`, but I don't know if that is the best approach. I would like to verify that the resulting package is importable and usable, but perhaps the preferred approach is to simply verify that the expected so and __init__.py files are present in the build output?

henryiii:
I'll try to review & help soon. Trying to prepare a tutorial submission for SciPy. :)

henryiii:
We've had this issue for a while. Scikit-build-core is careful in how it sets up tests, but scikit-build's older test suite has some issues with the host environment.

vyasr:
Got it, I'm guessing that since you merged you are OK with the `build_ext --inplace` approach to testing? I will note that it is not quite as good of a test as a true installation since there could be problems with things like install rules, but in practice all the issues I ran into were in the tests I wrote rather than in the main fix so the fix should be pretty robust.

PR #872, opened 2023-02-28
Provide more useful error if user provides CMAKE_INSTALL_PREFIX

### LINKED ISSUE #721: `CMAKE_INSTALL_PREFIX` breaks build_ext
Specifying the `CMAKE_INSTALL_PREFIX` with `build_ext` causes problems for scikit-build. The problem is due to the following properties of scikit-build:
- We parse and forward along all CMake arguments to CMake
- The `python setup.py install` command does not map to CMake installs, it instead uses standard setuptools and some Python functions for the copying required in the installation step.
- CMake `install` commands _are_ invoked as part of the build step to copy from `cmake-build` -> `cmake-install`
- Specifying `CMAKE_INSTALL_PREFIX` ends up affecting the `cmake-build` -> `cmake-install` copy

That last bit is what causes problems. If the install prefix is outside the source directory (which it usually is), we hit [this condition](https://github.com/scikit-build/scikit-build/blob/master/skbuild/setuptools_wrap.py#L796) and fail.

Without deeper knowledge of scikit-build, there are two options that I can think of:
- Either ignore or error out when `CMAKE_INSTALL_PREFIX` is specified. Do we actually need to support this variable for any purpose? Python installation always proceeds using `setup.py install` or `pip install`, and in both cases users only need to install to site-packages or the user local site-packages (using --user).
- Parse out `CMAKE_INSTALL_PREFIX` from the provided CMake arguments and then use it to determine the behavior of `python setup.py install`. This option would make sense if a goal of scikit-build is to support installation to some arbitrary directory similar to what CMake supports.

vyasr:
Resolves #721. I did not bother to add testing since this is just a helping the user understand an invalid edge case (and given the overall thrust towards scikit-build-core migration anyway).

### DIFF
--- a/skbuild/cmaker.py
+++ b/skbuild/cmaker.py
@@ -18,7 +18,6 @@ import shlex
 import subprocess
 import sys
 import sysconfig
-import warnings
 from pathlib import Path
 from shlex import quote
 from typing import Mapping, Sequence, overload
@@ -317,15 +316,8 @@ class CMaker:
             if any("CMAKE_INSTALL_PREFIX" in arg for arg in env_cmake_args):
                 raise ValueError("CMAKE_INSTALL_PREFIX may not be passed via SKBUILD_CONFIGURE_OPTIONS.")
         else:
-            env_cmake_args_filtered = list(filter(None, shlex.split(os.environ.get("CMAKE_ARGS", ""))))
+            env_cmake_args_filtered = filter(None, shlex.split(os.environ.get("CMAKE_ARGS", "")))
             env_cmake_args = [s for s in env_cmake_args_filtered if "CMAKE_INSTALL_PREFIX" not in s]
-            # Don't error, only warn in this instance because CMAKE_ARGS may be out of
-            # the project's control (e.g. when provided by conda builds).
-            if len(env_cmake_args) < len(env_cmake_args_filtered):
-                warnings.warn(
-                    "CMAKE_INSTALL_PREFIX was passed via CMAKE_ARGS and ignored. This "
-                    "cmake option is ignored and should be removed."
-                )
 
         cmd.extend(env_cmake_args)
 
--- a/skbuild/cmaker.py
+++ b/skbuild/cmaker.py
@@ -315,9 +315,7 @@ class CMaker:
         if "SKBUILD_CONFIGURE_OPTIONS" in os.environ:
             env_cmake_args = list(filter(None, shlex.split(os.environ["SKBUILD_CONFIGURE_OPTIONS"])))
             if any("CMAKE_INSTALL_PREFIX" in arg for arg in env_cmake_args):
-                raise ValueError(
-                    "CMAKE_INSTALL_PREFIX may not be passed via SKBUILD_CONFIGURE_OPTIONS."
-                )
+                raise ValueError("CMAKE_INSTALL_PREFIX may not be passed via SKBUILD_CONFIGURE_OPTIONS.")
         else:
             env_cmake_args_filtered = list(filter(None, shlex.split(os.environ.get("CMAKE_ARGS", ""))))
             env_cmake_args = [s for s in env_cmake_args_filtered if "CMAKE_INSTALL_PREFIX" not in s]
--- a/skbuild/cmaker.py
+++ b/skbuild/cmaker.py
@@ -18,6 +18,7 @@ import shlex
 import subprocess
 import sys
 import sysconfig
+import warnings
 from pathlib import Path
 from shlex import quote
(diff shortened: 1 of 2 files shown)


skbuild/cmaker.py
```diff
@@ -313,9 +314,18 @@ def configure(
...
+                raise ValueError("CMAKE_INSTALL_PREFIX may not be passed via SKBUILD_CONFIGURE_OPTIONS.")
         else:
-            env_cmake_args_filtered = filter(None, shlex.split(os.environ.get("CMAKE_ARGS", "")))
+            env_cmake_args_filtered = list(filter(None, shlex.split(os.environ.get("CMAKE_ARGS", ""))))
             env_cmake_args = [s for s in env_cmake_args_filtered if "CMAKE_INSTALL_PREFIX" not in s]
+            # Don't error, only warn in this instance because CMAKE_ARGS may be out of
+            # the project's control (e.g. when provided by conda builds).
+            if len(env_cmake_args) < len(env_cmake_args_filtered):
+                warnings.warn(
+                    "CMAKE_INSTALL_PREFIX was passed via CMAKE_ARGS and ignored. This "
```
henryiii, line 326: Is it important to warn? This is _very_ common, as it will always happen on conda builds.

vyasr: No, the warning isn't really necessary. I added it for consistency since it seemed odd to have some annotation for every other possible way of providing CMake arguments, but I can remove it.

skbuild/cmaker.py
```diff
@@ -18,6 +18,7 @@
 import subprocess
 import sys
 import sysconfig
+import warnings
```
vyasr, line 21:
```suggestion
```

skbuild/cmaker.py
```diff
@@ -313,9 +314,18 @@ def configure(
         # Parse CMAKE_ARGS only if SKBUILD_CONFIGURE_OPTIONS is not present
         if "SKBUILD_CONFIGURE_OPTIONS" in os.environ:
             env_cmake_args = list(filter(None, shlex.split(os.environ["SKBUILD_CONFIGURE_OPTIONS"])))
+            if any("CMAKE_INSTALL_PREFIX" in arg for arg in env_cmake_args):
+                raise ValueError("CMAKE_INSTALL_PREFIX may not be passed via SKBUILD_CONFIGURE_OPTIONS.")
         else:
-            env_cmake_args_filtered = filter(None, shlex.split(os.environ.get("CMAKE_ARGS", "")))
+            env_cmake_args_filtered = list(filter(None, shlex.split(os.environ.get("CMAKE_ARGS", ""))))
```
vyasr, line 320:
```suggestion
            env_cmake_args_filtered = filter(None, shlex.split(os.environ.get("CMAKE_ARGS", "")))
```

skbuild/cmaker.py
```diff
@@ -313,9 +314,18 @@ def configure(
...
             env_cmake_args = [s for s in env_cmake_args_filtered if "CMAKE_INSTALL_PREFIX" not in s]
+            # Don't error, only warn in this instance because CMAKE_ARGS may be out of
+            # the project's control (e.g. when provided by conda builds).
+            if len(env_cmake_args) < len(env_cmake_args_filtered):
+                warnings.warn(
+                    "CMAKE_INSTALL_PREFIX was passed via CMAKE_ARGS and ignored. This "
+                    "cmake option is ignored and should be removed."
+                )
```
vyasr, line 328:
```suggestion
```

skbuild/cmaker.py
henryiii: CMAKE_ARGS is a bit special, since it's picking up something general that might not be targeted to scikit-build. The other two are always targeted at scikit-build.

henryiii:
Thanks for the PRs!

vyasr:
np, trying to clean up some of my backlogged requests on this repo before moving my attention to scikit-build-core 😄

skbuild/cmaker.py
vyasr: Out of curiosity, does any tool use `CMAKE_ARGS` other than conda-forge builds? What concerns me a little here is the possibility that scikit-build is being developed for compatibility with conda-forge rather than the other way around. I think it's fine for scikit-build (which aims to be a reasonably drop-in replacement for setuptools) but I want to make sure that same mindset doesn't percolate to scikit-build-core. With modern PEP 517 builders, which could all be developed completely independently of one another, I think the responsibility should be on whatever invokes a build tool (e.g. conda-forge) to set up arguments in a way that is compatible with any builder that it wants to support.

PR #873, opened 2023-02-28
Match setuptools behavior for include_package_data default.

### LINKED ISSUE #864: Incompatibility with tool.setuptools.packages in pyproject.toml
I'll open by saying that I think it's completely fine if this is marked as a wontfix as out of scope for scikit-build, but it might be helpful to document the limitations of scikit-build, and possibly provide this as one reason to switch to scikit-build-core (perhaps this already exists and I just couldn't find it). For context, I found this while attempting to slim down setup.py files as much as possible prior to attempting a conversion from scikit-build to scikit-build-core.

setuptools has beta support for pyproject.toml files (the fact that it's beta has no bearing on the issue AFAICT). One of the features of this approach is using the `tools.setuptools` table to hold fields in addition to those documented in PEP 621 to extend functionality. In general, I suspect that many of these fields will work as expected if the operation is entirely contained within the setuptools invocation in scikit-build's `setuptools_wrap.setup`. However, some of these fields will not work as expected. The particular example that I ran into was the `tools.setuptools.packages` table. The problem with attempting to use this table is that there are steps upstream of the call to `setuptools.setup` in `setuptools_wrap.setup` that require the list of packages, and these steps will not receive the full list of packages since those are computed later by setuptools if they are specified using pyproject.toml. As a result, projects using scikit-build must still pass the package list as arguments in setup.py.

vyasr:
This PR attempts to match the default behavior of `include_package_data` in setuptools, which is somewhat subtle in that the default is dependent on the existence of certain tables in pyproject.toml.

Resolves #864.

### DIFF
--- a/skbuild/setuptools_wrap.py
+++ b/skbuild/setuptools_wrap.py
@@ -674,7 +674,20 @@ def setup(
     )
 
     original_manifestin_data_files = []
-    if kw.get("include_package_data", False):
+
+    def get_default_include_package_data():
+        # Include package data if pyproject.toml contains the project or tool.setuptools table.
+        # https://setuptools.pypa.io/en/latest/history.html#id255
+        # https://github.com/pypa/setuptools/pull/3067
+        pyproject_file = os.path.join(os.getcwd(), "pyproject.toml")
+        if os.path.isfile(pyproject_file):
+            with open(pyproject_file) as f:
+                text = f.read()
+                if "[project]" in text or "[tool.setuptools]" in text:
+                    return True
+        return False
+
+    if kw.get("include_package_data", get_default_include_package_data()):
         original_manifestin_data_files = parse_manifestin(os.path.join(os.getcwd(), "MANIFEST.in"))
         for path in original_manifestin_data_files:
             _classify_file(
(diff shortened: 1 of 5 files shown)


skbuild/setuptools_wrap.py
```diff
@@ -674,7 +674,20 @@ def setup(
...
+
+    def get_default_include_package_data() -> bool:
+        # Include package data if pyproject.toml contains the project or tool.setuptools table.
+        # https://setuptools.pypa.io/en/latest/history.html#id255
+        # https://github.com/pypa/setuptools/pull/3067
+        pyproject_file = os.path.join(os.getcwd(), "pyproject.toml")
+        if os.path.isfile(pyproject_file):
+            with open(pyproject_file) as f:
+                text = f.read()
+                if "[project]" in text or "[tool.setuptools]" in text:
```
henryiii, line 686: I'm okay to require tomli for Python < 3.11.

vyasr:
Thanks for making the change to use tomli/tomllib instead of manual text parsing! As you noted I wasn't sure if you wanted to introduce the dependency.

PR #879, opened 2023-03-06
tests: custom venv tester

henryiii:
Dropping usage of pytest-virtualenv, which is not well maintained and doesn't work everywhere.


.pre-commit-config.yaml
```diff
@@ -68,7 +68,7 @@ repos:
   rev: "v1.3.0"
   hooks:
   - id: mypy
-    files: ^(skbuild)  # TODO: add |tests
+    files: ^(skbuild|tests) # TODO: add |tests
```
henryiii, line 71:
```suggestion
    files: ^(skbuild|tests)
```

skbuild/setuptools_wrap.py
```diff
@@ -394,7 +394,7 @@ def setup(
     cmake_process_manifest_hook: Callable[[list[str]], list[str]] | None = None,
     cmake_install_target: str = "install",
     **kw: Any,
-) -> None:
+) -> upstream_Distribution:
```
henryiii, line 397: This is technically incorrect in typeshed.

PR #899, opened 2023-03-24
refactor: avoid mutating function input parameters

### LINKED ISSUE #749: Build type always ends up on default "Release"
Hi,
In https://github.com/scikit-build/scikit-build/blob/master/skbuild/setuptools_wrap.py#:~:text=%23%20Since%20CMake%20arguments,%5D%20%2B%20cmake_args
should
```
cmake_args = skbuild_kw["cmake_args"] + cmake_args
```
rather be
```
cmake_args = cmake_args + skbuild_kw["cmake_args"]
```
Otherwise, regardless of the content of environment variable `CMAKE_ARGS` or of the setup configuration:
```
skbuild.setup(
    ...
    cmake_args=['-DCMAKE_BUILD_TYPE:STRING=Abc123']
    ...
)
```
scikit-build always ends up with a `cmake_args` with a trailing `'-DCMAKE_BUILD_TYPE:STRING=Release'` (coming from https://github.com/scikit-build/scikit-build/blob/master/skbuild/setuptools_wrap.py#:~:text=%3DFalse)-,parser.add_argument(,),-parser.add_argument), hence discarding whatever build type the user is passing.

henryiii:
Was targeting #749, but that will be a followup now. This just changes parameter handling so that input parameters are no longer mutated. (yuck)

### DIFF
--- a/skbuild/setuptools_wrap.py
+++ b/skbuild/setuptools_wrap.py
@@ -523,11 +523,6 @@ def setup(
 
     data_files = {(parent_dir or "."): set(file_list) for parent_dir, file_list in kw.get("data_files", [])}
 
-    # Since CMake arguments provided through the command line have more
-    # weight and when CMake is given multiple times a argument, only the last
-    # one is considered, let's prepend the one provided in the setup call.
-    cmake_args = list(cmake_args) + cmake_args_from_args
-
     # Handle cmake_install_target
     # get the target (next item after '--install-target') or return '' if no --install-target
     cmake_install_target_from_command = next(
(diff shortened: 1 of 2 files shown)


henryiii:
We also need to be able to set `--config`, so this really isn't the right fix. The parsing functions append to mutable input arguments (!), so first probably need to fix that, otherwise this is going to be brittle and hard to change.

skbuild/setuptools_wrap.py
```diff
@@ -104,37 +104,40 @@ def parse_skbuild_args(
...
 
     # Consider CMake arguments passed as global setuptools options
-    cmake_args.extend([arg for arg in args if _is_cmake_configure_argument(arg)])
+    _cmake_args = [*cmake_args, *(arg for arg in args if _is_cmake_configure_argument(arg))]
     # ... and remove them from the list
-    args = [arg for arg in args if not _is_cmake_configure_argument(arg)]
+    _args = [arg for arg in args if not _is_cmake_configure_argument(arg)]
+    _build_tool_args = list(build_tool_args)
 
     namespace, remaining_args = parser.parse_known_args(args)
```
henryiii, line 112:
```suggestion
    namespace, remaining_args = parser.parse_known_args(_args)
```

henryiii:
This now is just fixing the mutating function parameters to make this easier to fix in the next PR.

Issue #900, opened 2023-03-26
cross-compiling from osx-64 to osx-arm64 can lead to badly linked libraries


erykoff:
This is a follow-up to https://github.com/scikit-build/scikit-build/issues/589 which was not fixed correctly by https://github.com/scikit-build/scikit-build/pull/622

The issue is that in cross-compiling `skip_test` is marked as `TRUE` here https://github.com/scikit-build/scikit-build/blob/f4f9796e8a940a5559264045b1c43105b291d6c2/skbuild/resources/cmake/targetLinkLibrariesWithDynamicLookup.cmake#L500-L502 which short-circuits here https://github.com/scikit-build/scikit-build/blob/f4f9796e8a940a5559264045b1c43105b291d6c2/skbuild/resources/cmake/targetLinkLibrariesWithDynamicLookup.cmake#L504-L506 and therefore @isuruf's patch https://github.com/scikit-build/scikit-build/blob/main/skbuild/resources/cmake/targetLinkLibrariesWithDynamicLookup.cmake#L411-L413 is never triggered.

Not sure of the best way to work around this...

henryiii:
scikit-build (classic) after #1185 is now a thin wrapper around scikit-build-core's setuptools plugin. Closing all open issues; if this still applies, please reopen or open an issue on scikit-build-core. Thanks!

PR #901, opened 2023-03-26
targetLinkLibrariesWithDynamicLookup: Fix support for cross-compilation exception

erykoff:
This pull-request fixes dynamic linking when cross-compiling on macOS.

It revisits the approach used to support cross-compilation when no emulator is available.

This is done by further improving on fixes integrated through https://github.com/scikit-build/scikit-build/commit/c0ec848644a22fb165018be7882b24099dcec2ae (`Fix targetLinkLibrariesWithDynamicLookup`) and https://github.com/scikit-build/scikit-build/commit/07b2c4ff0b03f830456eb1c17bae87437d5bf5c5 (`fix: assume success for osx_dynamic_lookup when cross compiling`)

Messages reported during configuration are also improved, when the compile tests are assumed to pass when cross-compiling without emulator, a clear message is reported.

For example:

```
Performing Test Weak Link MODULE -> SHARED (osx_dynamic_lookup) - Assuming Success (Cross compiling without emulator on macOS)
```

Fixes:
* #900
* https://github.com/conda-forge/h3-py-feedstock/issues/38

<details><summary>Previous description</summary>
This PR is a possible solution to #900, building on what @isuruf had done previously.  I don't think this would have any bad side effects, but I'm not very familiar with cmake so I could be wrong.

I did confirm that with this patch then `h3-py` cross-compiles correctly, so this would fix https://github.com/conda-forge/h3-py-feedstock/issues/38
</details>

### DIFF
--- a/skbuild/resources/cmake/targetLinkLibrariesWithDynamicLookup.cmake
+++ b/skbuild/resources/cmake/targetLinkLibrariesWithDynamicLookup.cmake
@@ -198,6 +198,28 @@ function(_test_weak_link_project
   set(osx_dynamic_lookup           "-undefined dynamic_lookup")
   set(no_flag                                               "")
 
+  if(CMAKE_CROSSCOMPILING)
+    set(link_flag_spec "no_flag")
+    set(link_flag "${${link_flag_spec}}")
+    set(test_skipping_reason "")
+    set(test_pass FALSE)
+
+    if(APPLE AND NOT CMAKE_CROSSCOMPILING_EMULATOR)
+      set(link_flag_spec "osx_dynamic_lookup")
+      set(link_flag "${${link_flag_spec}}")
+      set(test_skipping_reason " (Cross compiling without emulator on macOS)")
+      set(test_pass TRUE)
+    endif()
+
+    if(test_pass)
+      set(test_description "Weak Link ${target_type} -> ${lib_type} (${link_flag_spec})")
+      message(STATUS "Performing Test ${test_description} - Assuming Success${test_skipping_reason}")
+      set(${can_weak_link_var} ${test_pass} PARENT_SCOPE)
+      set(${project_name} ${link_flag} PARENT_SCOPE)
+      return()
+    endif()
+  endif()
+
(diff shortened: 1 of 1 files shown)


henryiii:
I think this is reasonable. @jcfr or @aaron-bray, let me know if you see any problems, otherwise I think it's safe to merge.

erykoff:
Thanks!  This does look cleaner than my hacked together patch.  I will test in my use case and report back.

skbuild/resources/cmake/targetLinkLibrariesWithDynamicLookup.cmake
```diff
@@ -198,6 +198,28 @@ function(_test_weak_link_project
...
 
+  if(CMAKE_CROSSCOMPILING)
+    set(link_flag_spec "no_flag")
+    set(link_flag "${link_flag_spec}")
+    set(test_skipping_reason "")
+    set(test_pass FALSE)
+
+    if(APPLE AND NOT CMAKE_CROSSCOMPILING_EMULATOR)
+      set(link_flag_spec "osx_dynamic_lookup")
+      set(link_flag "${link_flag_spec}")
```
erykoff, line 209: So this doesn't work, the cross-build errors with `error: no such file or directory: 'osx_dynamic_lookup'`.  I think that there's something here with a string replacement that isn't happening (it should be substituting `-undefined dynamic_lookup`) but something with quotes/brackets/?? is going wrong.

jcfr: Thanks for testing, the next commit should address this.

_Note that once we have confirmed it works, commits will be squashed_

erykoff:
With the latest fix, it works for my test cross compiling from osx-64 to osx-arm64.  Thanks!

henryiii (approved):

jcfr:
> With the latest fix, it works for my test cross compiling from osx-64 to osx-arm64. Thanks!

Great. I will update the commit message and this should be good for integration.

henryiii:
When you squash and merge, you can set the commit message there. ;)

PR #913, opened 2023-04-06
fix: include library for FindPython

### LINKED ISSUE #912: cross compiling for Windows Arm64
From the release notes there is experimental support to cross compile for Windows Arm64 using cibuildwheel since version [0.16.4](https://github.com/scikit-build/scikit-build/releases/tag/0.16.4). I did try to use this in my projects which fails with linker errors about missing symbols: https://github.com/maxbachmann/RapidFuzz/actions/runs/4615046555/jobs/8158606227. Disabling LTO leads to a different set of linker errors: https://github.com/maxbachmann/RapidFuzz/actions/runs/4630266637/jobs/8191700581. Is there anything in particular that need to be changed in projects for this to work?

henryiii:
Attempting to fix #912; I though this had a problem before, but let's try it.

### DIFF
--- a/skbuild/cmaker.py
+++ b/skbuild/cmaker.py
@@ -280,19 +280,24 @@ class CMaker:
             "-DSKBUILD:INTERNAL=TRUE",
             f"-DCMAKE_MODULE_PATH:PATH={cmake_resource_dir}",
             f"-DPYTHON_EXECUTABLE:PATH={sys.executable}",
-            f"-DPYTHON_INCLUDE_DIR:PATH={python_include_dir}",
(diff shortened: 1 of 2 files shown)


henryiii:
This seems to not break anything, @maxbachmann would it be possible to see if it fixes the issue you are seeing cross compiling with FindPython?

maxbachmann:
Yes this works as expected: https://github.com/maxbachmann/RapidFuzz/actions/runs/4635534506/jobs/8202692945

henryiii:
Thanks! We are probably pretty close to a release.

PR #917, opened 2023-04-09
Fix cross-compiling Windows ARM64

dlech:
The code that creates the `default_generators` for Visual Studio generators was using the host architecture instead of the target architecture when cross-compiling so finding the compiler would fail.

This reuses the existing helper function to get the architecture for the candidate generators.


henryiii:
Thanks!

Issue #940, opened 2023-04-15
setting Python3_LIBRARY is broken in virtual environments in MSYS2


dlech:
I tried building a project with scikit-build 0.17.1 using `python -m build` which builds in a virtual environment in mingw on windows. It fails while configuring cmake. Here is the relevant output.

```
Configuring Project
  Working directory:
    C:/msys64/tmp/build-via-sdist-oo6vlkz6/winsdk-1.0.0b8/_skbuild/mingw_aarch64-3.10/cmake-build
  Command:
[...]

PR #941, opened 2023-04-15
fix: Don't set Python_LIBRARY when using venv

dlech:
If scikit-build is running in a virtual environment, then the path to libpython is not available. The virtual environment does not copy this file and it overrides sysconfig variables so we can no longer find the original location of the library.

Fixes: https://github.com/scikit-build/scikit-build/issues/940

### DIFF
--- a/skbuild/cmaker.py
+++ b/skbuild/cmaker.py
@@ -504,6 +504,12 @@ class CMaker:
                 minor = sys.version_info[1]
                 return str(Path(result) / f"python3{minor}.lib")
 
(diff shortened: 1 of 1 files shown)


dlech:
I suppose the issue I reported is specific to mingw. Clearly this does not seem to be the correct solution.

henryiii:
I think we need to only pass the lib for MSVC windows. Passing it breaks FindPython still. This is a bug in FindPython that wasn’t fully fixed in 3.24, it seems.

Issue #957, opened 2023-04-26
Can't find python library for statically linked distribution


tommy-waltmann:
Hi scikit-build team!

I maintain an open-source package for particle trajectory analysis, [freud](https://github.com/glotzerlab/freud), which uses the `scikit-build` tool. We build and test wheels of our package regularly with github actions on publicly available manylinux images. Some of those images use statically linked python distributions instead of the much more common dynamically linked distributions.

The most recent `scikit-build` release, 0.17.2, is failing to find the python library in those images and the build fails. I have opened a draft PR in freud [here](https://github.com/glotzerlab/freud/pull/1098) to demonstrate the issues we are having. The failing `build_wheels` test is available so the errors, environment, and steps to reproduce can be inspected more closely.

Relevant environment details:
manylinux image: [quay.io/pypa/manylinux2014_x86_64:2023-03-05-271004f](http://quay.io/pypa/manylinux2014_x86_64:2023-03-05-271004f)
python version: >=3.7
scikit-build version: 0.17.2

I can contribute code for a PR to fix this, but will need direction from maintainers on what to change.

henryiii:
Are you sure it used to work? There are no static libs in the manylinux images. They are stripped out to save space and avoid users incorrectly linking to them when building modules.

The only platform you need to link on is Windows.

henryiii:
0.17.2 no longer adds PYTHON_LIBRARIES if it would point at a non-existing file. This makes FindPythonLibs unhappy. But it’s not needed in this case, so I think that’s the correct fix (please test! My analysis could be wrong). This might have further implications, though, so will have to see. It’s easy to work around if you had to, though, just set PYTHON_LIBRARIES to a fake location if it’s not defined. 🤦

Python modules do not need the Python symbols - Python already had the symbols loaded, so there’s no need to link to it, and doing so could possibly cause you to break on patch releases. So it’s not allowed. (Not true on Windows, where it does something in setting up the DLL).

If you build an app that embeds Python in it, you need to link; manylinux doesn’t support that, though.

FrancescAlted:
[...]

PR #958, opened 2023-04-26
fix: FindNumPy should call FindPythonLibs

henryiii:
This should not be necessary. It should already be found, and even if not, PYTHON_EXECUTABLE is set and PYTHON_INCLUDE_DIR isn't important. I think this is causing #957, since you can't run FindPythonLibs REQUIRED on manylinux.


henryiii:
That warning is correct; on manylinux, there are no Python libs. Your users are not likely to be building on manylinux (since it's only for making wheels), so will probably not get that warning - most Python distributions do include the libs. The root problem I expect is PythonExtensions is calling FindPythonLibs, which triggers that warning. Maybe we can see if we can fix that.

PR #959, opened 2023-04-27
revert: keep return type string if python lib non-existing for now

henryiii:
Suggested improvement for openCV. I think this should go back to returning None in 0.18, but this would help the perceived regression for OpenCV in 0.17.2. See #940.


henryiii:
Another option would be to return a dummy string, which would keep the old behavior exactly and then go back to None in 0.18. I think we can try this one first, though.

PR #983, opened 2023-05-17
fix: add support for sunos

### LINKED ISSUE #520: Ask for Solaris or illumos support
Hi

I am running Solaris 11.4. It looks like `skbuild/platform_specifics/platform_factory.py` is stopping the compilation of the scikit build on Solaris.

I see the documentation calls out support for MacOS/Windows/Linux. But I do see the code in platform_factory.py supports bsd and os400. 

Given bsd can get this to work I believe I could get it work on solaris. Any chance sunos could be added platform_factory.py so I could give it go? 

Thanks for your consideration!

JIM

mtelka:
Fixes #520

### DIFF
--- a/skbuild/platform_specifics/platform_factory.py
+++ b/skbuild/platform_specifics/platform_factory.py
@@ -40,5 +40,10 @@ def get_platform() -> abstract.CMakePlatform:
 
         return bsd.BSDPlatform()
 
+    if this_platform == "sunos":
+        from . import sunos
+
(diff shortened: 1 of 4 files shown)


henryiii (approved):

henryiii:
Thank you! I’m going to have to take your word for it that it works. :) (Though looks like there may be a GHA that will set up a VM for some of these operating systems.)

PR #990, opened 2023-05-27
tests: fix match for get_map symbol

keszybz:
Wrong line could be matched by mistake, e.g.
    45: 0000000000010160     0 NOTYPE  LOCAL  DEFAULT   10 0000001a.plt_call.get_map


henryiii:
3.12 failures are due to https://github.com/man-group/pytest-plugins/issues/220. #879 would be a way forward, possibly.

henryiii:
Thanks!

PR #996, opened 2023-06-08
Add note on install_manifest.txt to usage.rst

ihnorton:
Add a note on how skbuild uses install_manifest, and warn that EPs can be problematic (no install_manifest in the expected directory -> no module in the wheel). #997

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -413,7 +413,7 @@ How does scikit-build know where to find compiled modules?
 ----------------------------------------------------------
 
 Compiled targets and other install files are identified by parsing the CMake-generated
-``install_manifest.txt`` file from :func:`skbuild.constants.CMAKE_BUILD_DIR()`. Note:
+``install_manifest*.txt`` files from :func:`skbuild.constants.CMAKE_BUILD_DIR()`. Note:
 if scikit-build is driving nested CMake projects (externalproject / "superbuild"), the
 manifest file may not be present in the ``CMAKE_BUILD_DIR``.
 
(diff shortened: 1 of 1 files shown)


henryiii:
What does the install_manifest look like in super builds? Is it in subdirectories?

ihnorton:
>  Is it in subdirectories?

Yes (related: https://github.com/orgs/scikit-build/discussions/997)

docs/usage.rst
```diff
@@ -408,6 +408,15 @@ and a python wheel, it is possible to test for the variable ``SKBUILD``:
       message(STATUS "The project is built using scikit-build")
     endif()
 
+
+How does scikit-build know where to find compiled modules?
+----------------------------------------------------------
+
+Compiled targets and other install files are identified by parsing the CMake-generated
+``install_manifest.txt`` file from :func:`skbuild.constants.CMAKE_BUILD_DIR()`. Note:
```
jcfr, line 416:
```suggestion
``install_manifest*.txt`` files from :func:`skbuild.constants.CMAKE_BUILD_DIR()`. Note:
```

docs/usage.rst
```diff
@@ -408,6 +408,15 @@ and a python wheel, it is possible to test for the variable ``SKBUILD``:
...
+if scikit-build is driving nested CMake projects (externalproject / "superbuild"), the
+manifest file may not be present in the ``CMAKE_BUILD_DIR``.
```
jcfr, line 418: The way to currently deal with this is to explicitly include the `cmake_install.cmake` by adding code like the following in the outer project:

```
#-----------------------------------------------------------------------------
# Install inner project
message(STATUS "Adding install rules for inner project")
install(CODE "
include(\"${CMAKE_BINARY_DIR}/inner-build/cmake_install.cmake\")
")
```

Notes:
* Make sure to tweak `${CMAKE_BINARY_DIR}/inner-build` based on your project.
* It may be possible to generalize this to include `cmake_install.cmake` from other projects also built using `ExternalProject` ...

henryiii:

PR #1019, opened 2023-08-02
Avoid `PythonLibs` and `PythonInterp` due to CMP0148

### LINKED ISSUE #1018: CMP0148 with CMake version >= 3.27
I am currently running CMake 3.27 (release on July 18th; it is the default of CMake on openSUSE Tumbleweed).

However, if use the following:

```cmake
cmake_minimum_required(VERSION 3.26)
project(Test)
find_package(
  Python 3 EXACT
  COMPONENTS Interpreter
  REQUIRED
)

find_package(PythonExtensions REQUIRED)
```

and do the following:

```sh
python3 -m venv venv
source venv/bin/activate
pip install git+https://github.com/scikit-build/scikit-build
mkdir build
cd build
rm -rf * && cmake -DCMAKE_MODULE_PATH=$(readlink -f ../venv/lib64/python3.11/site-packages/skbuild/resources/cmake) ..
```

then I get the following:

```
-- Found Python: /mnt/baranem/cmake/venv/bin/python3.11 (found suitable exact version "3.11.4") found components: Interpreter
CMake Warning (dev) at venv/lib/python3.11/site-packages/skbuild/resources/cmake/FindPythonExtensions.cmake:245 (find_package):
  Policy CMP0148 is not set: The FindPythonInterp and FindPythonLibs modules
  are removed.  Run "cmake --help-policy CMP0148" for policy details.  Use
  the cmake_policy command to set the policy and suppress this warning.

Call Stack (most recent call first):
  CMakeLists.txt:9 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Found PythonInterp: /mnt/baranem/cmake/venv/bin/python3.11 (found version "3.11.4")
CMake Warning (dev) at venv/lib/python3.11/site-packages/skbuild/resources/cmake/FindPythonExtensions.cmake:252 (find_package):
  Policy CMP0148 is not set: The FindPythonInterp and FindPythonLibs modules
  are removed.  Run "cmake --help-policy CMP0148" for policy details.  Use
  the cmake_policy command to set the policy and suppress this warning.

Call Stack (most recent call first):
  CMakeLists.txt:9 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Found PythonLibs: /usr/lib64/libpython3.10.so (found version "3.10.12")
```

This is to do with https://cmake.org/cmake/help/latest/policy/CMP0148.html.

aytey:
This PR aims to avoid the use of `PythonLibs` and `PythonInterp` due to [CMP0148](https://cmake.org/cmake/help/latest/policy/CMP0148.html), which was introduced in CMake 3.27.

Closes #1018.

### DIFF
--- a/skbuild/resources/cmake/FindPythonExtensions.cmake
+++ b/skbuild/resources/cmake/FindPythonExtensions.cmake
@@ -242,14 +242,14 @@
 # limitations under the License.
 #=============================================================================
 
-find_package(PythonInterp REQUIRED)
+find_package(Python COMPONENTS Interpreter)
 if(SKBUILD AND NOT PYTHON_LIBRARY)
   set(PYTHON_LIBRARY "no-library-required")
-  find_package(PythonLibs)
+  find_package(Python COMPONENTS Development)
   unset(PYTHON_LIBRARY)
   unset(PYTHON_LIBRARIES)
 else()
-  find_package(PythonLibs)
+  find_package(Python COMPONENTS Development)
 endif()
 include(targetLinkLibrariesWithDynamicLookup)
 
@@ -297,7 +297,7 @@ sys.stdout.write(\";\".join((
 )))
 ")
 
-execute_process(COMMAND "${PYTHON_EXECUTABLE}" -c "${_command}"
+execute_process(COMMAND "${Python_EXECUTABLE}" -c "${_command}"
                 OUTPUT_VARIABLE _list
                 RESULT_VARIABLE _result)
 


aytey:
As a note for something not in the description: there are other uses of `PYTHON_EXECUTABLE` in the scikit-build source tree, *but* there are no other instances of `PythonLibs`/`PythonInterp`.

I am not sure if these other uses of `PYTHON_EXECUTABLE` need to be fixed (and if it is simple as simply `s/PYTHON_EXECUTABLE/Python_EXECUTABLE/g`).

henryiii:
If we just hard switch, this will break all downstream packages that use the variables set by FindPythonLibs/FindPythonInterp. What I'd do if we really want to save FindPythonExtensions is have it respect FindPython and not trigger FindPythonLibs/Interp if FindPython has already run. That's how pybind11 handles this.

Longer term, scikit-build-core doesn't have "FindPythonExtensions". If you use FindPython, there's very little if any need to use FindPythonExtensions. It was only needed to fill in missing parts of FindPythonLibs/FindPythonInterp.

aytey:
> If we just hard switch, this will break all downstream packages that use the variables set by FindPythonLibs/FindPythonInterp
>

Yep! I totally agree ... when I started needing to make any changes to the external visibility of `PYTHON_EXECUTABLE`, I knew this was going to be a "breaking change" for anyone that uses scikit-build.

> What I'd do if we really want to save FindPythonExtensions
>

I guess this is a question for you, and not one I can answer: do you?

> not trigger FindPythonLibs/Interp if FindPython has already run
>

Seems like a very reasonable change.

> That's how pybind11 handles this.
>

I can take a look at this and see if I can bring this change across to scikit-build.

> If you use FindPython, there's very little if any need to use FindPythonExtensions
>

Maybe I'm using it wrong then: I was using FindPythonExtensions to get `python_extension_module` and `python_standalone_executable` -- maybe I just don't need these?

henryiii:
For FindPython, you can use [`python_add_library`](https://cmake.org/cmake/help/latest/module/FindPython.html#commands).

aytey:
Yeah, it seems that:

```cmake
    add_library(${module} MODULE ${module})
    python_extension_module(${module})
```

Has the same result as:

```cmake
    Python_add_library(${module} MODULE WITH_SOABI ${module})
```

However, `python_standalone_executable` does add some target properties that I'd need to do manually if I didn't use it.

henryiii:
```cmake
add_executable(myexe ...)
target_link_libraries(myexe PRIVATE Python::Python)
```
?

hmaarrfk:
Just chiming in as a maintainer helping a few people cross compile.

>  It will also require a really new CMake version

I really wonder if a newer version of cmake would be in the spirit of SPEC0. All in all, those that are using old software likely in maintenance mode, and don't really need to update any of them, let alone something like scikit build.

Depending on how far you want to go:

* CMake 3.22.1 would give you support for "24 months" https://github.com/Kitware/CMake/releases/tag/v3.22.1 
* Cmake 3.19.2 would give you support for "36 months" https://github.com/Kitware/CMake/releases/tag/v3.19.2

I think that `FindPython` from Cmake 3.12 is thus "not really that new" and may warrant some breaking changes and a bit of cleanup. Looking at the documentation, https://cmake.org/cmake/help/latest/module/FindPython.html, it seems that most features are available since 3.19 (with the notable exception of SOABI_MODULE...) so I think it would come as a great boom.

Ultimately, cross compilation is becoming more and more important with new architectures poping up every year. 

I think some breaking changes might be necessary in order to help push things forward.

https://scientific-python.org/specs/spec-0000/

henryiii:

PR #1021, opened 2023-08-10
chore: use 2x faster black mirror

henryiii:
Committed via https://github.com/asottile/all-repos


jcfr:
For reference:
* https://black.readthedocs.io/en/latest/integrations/source_version_control.html
* https://github.com/psf/black/issues/3405
* https://github.com/psf/black/pull/3828

PR #1030, opened 2023-09-19
Use first available option for vswhere output

ZzEeKkAa:
If multiple locations are returned next line check is failing since there is no folder that is a concatenation of two paths. That results in multiple installs for the same VS version treated as there is no one installed.

### DIFF
--- a/skbuild/platform_specifics/windows.py
+++ b/skbuild/platform_specifics/windows.py
@@ -139,24 +139,33 @@ def _find_visual_studio_2017_or_newer(vs_version: int) -> str:
         return ""
 
     try:
-        path = subprocess.run(
-            [
-                os.path.join(root, "Microsoft Visual Studio", "Installer", "vswhere.exe"),
-                "-version",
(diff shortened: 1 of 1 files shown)


skbuild/platform_specifics/windows.py
```diff
@@ -139,24 +139,33 @@ def _find_visual_studio_2017_or_newer(vs_version: int) -> str:
...
-                "-products",
-                "*",
-            ],
-            encoding="utf-8" if sys.platform.startswith("cygwin") else "mbcs",
-            check=True,
-            stdout=subprocess.PIPE,
-            errors="strict",
-        ).stdout.strip()
+        # vswhere.exe may return multiple locations separated by new line. For
+        # example for 2022 Build Tools and for 2022 Community Edition. Idealy
```
henryiii, line 143:
```suggestion
        # example for 2022 Build Tools and for 2022 Community Edition. Ideally
```

Issue #723, opened 2022-06-06
How to append, rather than overwrite, the `CMAKE_MODULE_PATH` that skbuild determines?


Helveg:
I have a binary extension which ships its own CMake modules, and they aren't detected because `scikit-build` sets `CMAKE_MODULE_PATH` itself to a specific absolute path. Is there an argument that allows me to append paths to this? My current workaround is to pass `-- -DCMAKE_MODULE_PATH=<my_path>:<skbuild_path>` but it requires me to manually check what path `scikit-build` uses on that machine/Python-env. It doesn't seem to be documented anywhere and seems like a generic enough issue to document a workaround for.

henryiii:
scikit-build (classic) after #1185 is now a thin wrapper around scikit-build-core's setuptools plugin. Closing all open issues; if this still applies, please reopen or open an issue on scikit-build-core. Thanks!

PR #1039, opened 2023-11-12
Prepend to CMAKE_MODULE_PATH if it already exists

opoplawski:
Possible fix for #723 although it depends on setting an environment variable instead of just picking up CMAKE variables, though I'm not sure that is possible.


henryiii:

PR #1052, opened 2023-12-16
Try to remove old configuration code that is required for numpy 0.15

hmaarrfk:


henryiii:
I think this can be swapped for something else, not directly removed. The note mentions entry points.

hmaarrfk:
honestly, I'm just not sure I really have the bandwidth to hit the spread in use cases currently required by scitkit build.

I personally think the compatibility matrix needs to be narrowed and more closely aligned with SPEC 0.

if you are open to that, I can probably dedicate a bit more time, but the current state of things makes it really hard to be a part-time contributor to this project.

PR #1101, opened 2024-07-01
iOS and Android support

FeodorFitsner:
This modification enables the execution of scikit-build within a Python crossenv environment, catering to both iOS and Android platforms.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,8 @@ build-backend = "hatchling.build"
 
 [project]
 name = "scikit-build"
-dynamic = ["version", "readme"]
+version = "0.17.6"
+dynamic = ["readme"]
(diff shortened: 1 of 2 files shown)


henryiii:
Thanks!

henryiii:
Is there any way to test in CI, by the way?

FeodorFitsner:
It depends on what we would like to test. If it's building a wheel using scikit-build then the process includes:

1) compile Python for the target platform (iOS or/and Android). For iOS it should be run on macOS CI worker. It's around 10-20 minutes.
2) install `crossenv` Python package and create new virtual environment.
3) run `build` command in that environment to build a wheel of a package using scikit-build.

If we need to test created iOS/Android wheel then it's additionally:'

4) create iOS/Android app with Python runtime and created wheel.
5) run the app in simulator.

PR #1108, opened 2024-07-31
test: remove test directives

s-t-e-v-e-n-k:
setuptools 72 and above now raise warnings for test_suite and tests_require directives in setup.py as support for setup.py test is preparing to be removed. These directives have been deprecated for quite some time, so just remove them from the sample setup.py files.


henryiii:
Windows tests stopped with KeyboardInterrupt? That's odd.

henryiii:
Thanks! This isn't causing the Windows timeout so will work on that separately. Later. :)

Issue #449, opened 2019-09-22
CMake finds wrong F2PY


bnavigator:
I know [CMake suggests](https://cmake.org/cmake/help/latest/command/find_program.html) to put the unversioned filename in front for the `find_program()` call:

> NAMES
> Specify one or more possible names for the program.
> 
> When using this to specify names with and without a version suffix, we recommend specifying the unversioned name first so that locally-built packages can be found before those provided by distributions.

But in the case of python files this is counterproductive. When you want to build a project for Python 2 where both Python 2 and 3 are installed, this call always finds the default `f2py`:

https://github.com/scikit-build/scikit-build/blob/5b20908709070e6329348aabbb889bacc25ca892/skbuild/resources/cmake/FindF2PY.cmake#L69

At python-control/Slycot#81 this lead to a hard to track down `free(): invalid pointer` error in the python2 packaging for Arch Linux and OpenSUSE, because the wrong f2py was used. As a workaround one can provide the path to the correct executable by adding the variable explicitly during the build call:
`python2 setup.py build -- -DF2PY_EXECUTABLE=/usr/bin/f2py2`

I suggest putting the unversioned filename to the end of the `find_program()` call and additionally making sure that the f2py is from the correct numpy.

jcfr:
Thanks for the suggestion :pray: , makes sense. I am integrating the corresponding change now. I will also make sure to give you credit for the contribution :+1:

PR #1111, opened 2024-08-06
FindF2PY: ensure generic f2py executable is looked up first

smiet:
In issue #449  a change was made to the resolution of F2py to support python2 builds and prefer the versioned `f2py` over an unversioned. 

https://github.com/scikit-build/scikit-build/blob/5b20908709070e6329348aabbb889bacc25ca892/skbuild/resources/cmake/FindF2PY.cmake#L69

This made sense at the time, but with the now final deprecation of python2, and numpy dropping `f2py3` in version [1.26](https://github.com/numpy/numpy/pull/24235) this change has come to bite us. 

Now if using a virtual environment with a version of `numpy>1.26`, but an OS/system that has a python with `numpy<1.26`, cmake will prefer the `f2py3` of the OS, leading to hard-to-diagnose problems. 

With `numpy>=2.0`  [C-API changes](https://numpy.org/devdocs/numpy_2_0_migration_guide.html#c-api-changes),  code compiled with the systems f2py will not run in the environment.

### DIFF
--- a/skbuild/resources/cmake/FindF2PY.cmake
+++ b/skbuild/resources/cmake/FindF2PY.cmake
@@ -66,7 +66,7 @@
 #   case, CMake is not used to find the compiler and configure the associated build system.
 #
 
-find_program(F2PY_EXECUTABLE NAMES f2py${PYTHON_VERSION_MAJOR} f2py)
+find_program(F2PY_EXECUTABLE NAMES f2py f2py${PYTHON_VERSION_MAJOR} )
 
 # XXX This is required to support NumPy < v0.15.0. See note in module documentation above.
 if(NOT F2PY_EXECUTABLE)


sdrave:
We have been affected by this issue as well. Trying to pip install [python-control/slycot](https://github.com/python-control/Slycot) in a venv will fail on any machine with a globally installed numpy<2. pip will pull numpy 2 as a build dependency and use its headers, but cmake will select the system f2py to generate the bindings.

smiet:
A quick and dirty workaround is to add a symlink called `f2py3` in your virtual environment:
`$ ln -s /path/to/venv/bin/f2py /path/to/venv/bin/f2py3`
Then CMake will look there first and not find the system's executable.

PR #1126, opened 2024-11-12
`setuptools_wrap.py`: parse `CMAKE_ARGS` with `shlex.split` like elsewhere

haampie:
* Fix a bug where `CMAKE_ARGS` was not parsed with `shlex.split`, which was breaking with e.g. `CMAKE_ARGS` set to `-G "Unix Makefiles"`.
* Remove redundant `list(filter(None, ...))`

I wish I could just pass the CMake arguments as command line arguments. Apparently calling `setup.py` is the only way to do so, but deprecated (?). What's left is patching toml files (not great), or setting this env variable (ugly).


henryiii:
What TOML file would be patched? Wasn't aware that scikit-build (classic) was affected by TOML files.

henryiii:
Thanks!

PR #1159, opened 2025-09-17
Fix failure with CMake 4

Flamefire:
CMake 4 removed compatibility with CMake 3.5 so no valid generator will be found as the test CMakeLists.txt causes an error.

Use the current CMake version to set the required version to avoid this

Current CMake version used is 4.1. The error looks like:
```
    -- Trying "Ninja" generator
    --------------------------------
    ---------------------------
    ----------------------
    -----------------
    ------------
    -------
    --
    Not searching for unused variables given on the command line.
    CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
      Compatibility with CMake < 3.5 has been removed from CMake.
    
      Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
      to tell CMake that the project requires at least <min> but has been updated
      to work with policies introduced by <max> or earlier.
    
      Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
    
    
    -- Configuring incomplete, errors occurred!
    --
    -------
    ------------
    -----------------
    ----------------------
    ---------------------------
    --------------------------------
    -- Trying "Ninja" generator - failure
...
    scikit-build could not get a working generator for your system. Aborting build.
```

BTW: Is there an easy way for a user to tell scikit-build to use an installed CMake instead of downloading one? That would also resolve this as the user can choose a lower version

### DIFF
--- a/skbuild/cmaker.py
+++ b/skbuild/cmaker.py
@@ -121,7 +121,7 @@ class CMaker:
         >>> open(cmake_fpath, 'w').write(ub.codeblock(
                 '''
                 message(STATUS "Using CMake ${CMAKE_VERSION}")
-                cmake_minimum_required(VERSION 3.5...3.20)
+                cmake_minimum_required(VERSION 3.5...3.26)
                 project(foobar NONE)
                 file(WRITE "${CMAKE_BINARY_DIR}/foo.txt" "# foo")
                 install(FILES "${CMAKE_BINARY_DIR}/foo.txt" DESTINATION ".")


emilazy:
Ah, I guess perhaps this fixes something about the build system orchestration itself, which isn’t expecting the CMake warning, without affecting the tests? I suppose we are bypassing that layer in Nixpkgs so we haven’t noticed it…

Both PRs might be needed, then.

Flamefire:
I added 2 more commits :

The first shows the CMake version used. With this the output will start with:
```
    Using CMake 4.1
    Not searching for unused variables given on the command line.
    CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
      Compatibility with CMake < 3.5 has been removed from CMake.
```

The other is just a minor doc enhancement to show using `3.5...3.20` CMake version ranges to keep the example working when `<3.10` is dropped

skbuild/platform_specifics/abstract.py
```diff
@@ -49,7 +49,8 @@ def write_test_cmakelist(languages: Iterable[str]) -> None:
         if not os.path.exists(test_folder):
             os.makedirs(test_folder)
         with open(f"{test_folder}/CMakeLists.txt", "w", encoding="utf-8") as f:
-            f.write("cmake_minimum_required(VERSION 3.5)\n")
+            f.write('message(STATUS "Using CMake ${CMAKE_VERSION}")\n')
+            f.write("cmake_minimum_required(VERSION ${CMAKE_VERSION})\n")
```
henryiii, line 53:
```suggestion
            f.write("cmake_minimum_required(VERSION 3.5...3.26)\n")
```

henryiii:
This was updated in #1166 along with lots of other fixes to get the CI working again (broken CI had things stuck for a while). Thanks, though!

PR #1160, opened 2025-09-17
tests: fix with CMake 4

emilazy:
CMake 4 dropped support for old policy versions and changed an error message.


tests/samples/issue-335-support-cmake-source-dir/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.2)
+cmake_minimum_required(VERSION 3.2...3.10)
```
henryiii, line 1:
```suggestion
cmake_minimum_required(VERSION 3.2...4.0)
```

PR #1185, opened 2026-06-09
feat!: use scikit-build-core's setuptools plugin as the backend

henryiii:
Produced with Claude Fable 5.

This replaces the classic scikit-build backend with scikit-build-core's setuptools plugin. `skbuild.setup()` is now a thin wrapper around `scikit_build_core.setuptools.wrapper.setup()`, and the legacy internals (`setuptools_wrap`, `cmaker`, `command/`, `platform_specifics/`, `utils/`, `_compat/` — about 3700 lines) are removed. Most packages, which only use `from skbuild import setup` with the documented `cmake_*` keywords, build unchanged.

Depends on scikit-build-core 1.0, now released: the runtime pin is `scikit-build-core[setuptools]>=1.0` from PyPI. For testing against an unreleased core, `SKBUILD_CORE_REQ` overrides the requirement in noxfile.py and tests/conftest.py.

### What still works
- `from skbuild import setup` with `cmake_args`, `cmake_source_dir`, `cmake_install_dir`, `cmake_install_target`, `cmake_process_manifest_hook`
- The shipped CMake modules (`FindPythonExtensions`, `FindCython`, `FindF2PY`, `FindNumPy`, ...), now injected via scikit-build-core's `cmake.module` entry-point group
- `SKBUILD` CMake variable (now `"2"`; still truthy), `setup.py build_ext --inplace`, standard setuptools commands
- `skbuild.exceptions.SKBuildError` (now an alias of setuptools' `SetupError`)

### Breaking changes (full notes in CHANGES.md)
- scikit-build-specific CLI options removed (`--skip-cmake`, `-G`, `--build-type`, `-j`, `--hide-listing`, the triple-`--` syntax, ...) — use `CMAKE_ARGS`/`CMAKE_GENERATOR`, the `cmake_args` keyword, or `[tool.scikit-build]`/`SKBUILD_*` settings
- `cmake_with_sdist=True` errors; `cmake_languages`/`cmake_minimum_required_version` warn and are ignored
- Internal modules (`skbuild.cmaker`, `skbuild.constants`, ...) removed
- `_skbuild/` directory gone (standard setuptools `build/` dirs); sdist manifests are no longer git-derived (use `MANIFEST.in`); editable installs need `editable.mode = "inplace"`
- Generator probing (incl. Visual Studio discovery) removed; CMake defaults apply

### Testing
- Suite triaged: internals tests deleted, end-to-end tests adapted; new error-path tests for the unsupported keywords.
- New `nox -s sample_projects` session + 3-OS CI job building scikit-build-sample-projects (checkout pinned to a fixed ref) against this checkout; all six classic-skbuild samples build (hello-cmake-package skipped on Windows, where its shared library exports no symbols).
- Compatibility gaps found by this suite were fixed in scikit-build-core before 1.0 (missing pyproject.toml, pure projects, `build_ext --inplace`, `exclude_package_data`, `package_dir` mapping, top-level py_modules, symlinked-cwd manifests).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


henryiii:
@LecrisUT thoughts on the Fedora side (1.0 is not there yet)?

LecrisUT:
> @LecrisUT thoughts on the Fedora side (1.0 is not there yet)?

Nope not yet, I'm looking into [it](https://src.fedoraproject.org/rpms/python-scikit-build-core/pull-request/85) today, need to do reverse dependency tests, check if setuptools optional dependency is exported, etc.

Branched Fedora will also not get this update so
https://github.com/scikit-build/scikit-build/blob/c116e8f5ebd570b23e5548dc987820fb14463d7e/.packit.yaml#L27-L28
should be changed to rawhide with a note to gradually re-enable.

LecrisUT:
Ok, rawhide update will come tomorrow. Should make this PR depend on https://github.com/scikit-build/scikit-build/pull/1201

henryiii:
Okay, let me know when I should rebase.

tests/test_include_exclude_data.py
```diff
@@ -2,12 +2,18 @@
...
 
 import pytest
+from packaging.version import Version
 
 from . import to_unix_path
 from .pytest_helpers import check_sdist_content, check_wheel_content
 
+# scikit-build-core >=1.0.3 reports CMake inputs via get_source_files(), so
+# sdists automatically include them.
+CORE_SDIST_HAS_CMAKE_SOURCES = Version(metadata.version("scikit-build-core")) >= Version("1.0.3")
```
LecrisUT, line 15: This is only needed because it's not on Fedora yet? I can unblock it today

henryiii: That's exactly why. Though allowing the tests to work with any 1+ instead of 1.0.3+ only isn't too bad.

henryiii:

Downstream A/B testing (released scikit-build vs this branch) found one packaging behavior change to look into later:

**Wheels no longer materialize shared-library symlinks.** Classic scikit-build copied each symlink into the wheel as a full duplicate file; the core backend ships only the real file. Seen with python-blosc on macOS arm64: the baseline wheel has `libblosc.1.21.7.dylib` plus full copies `libblosc.1.dylib` and `libblosc.dylib` under `.data/data/lib/`, while this branch ships only `libblosc.1.21.7.dylib` (wheel size 1.84 MB → 1.10 MB). python-blosc itself works fine either way, but a consumer linking against the unversioned name from the wheel data dir would break. Probably fine (arguably an improvement), but worth a changelog note.

Everything else in the batch (picologging, ckdl, confseq, python-blosc; line_profiler fails identically on both sides for project-side reasons) built with no regressions — only the expected `Generator` field change and the better deployment-target macOS platform tag.

henryiii:

Second downstream A/B batch (released scikit-build vs this branch). **One real incompatibility found (DracoPy); everything else OK.**

### DracoPy — build breaks on removed public modules

seung-lab/DracoPy builds fine with released scikit-build but fails immediately on this branch, before any CMake runs:

```
ModuleNotFoundError: No module named 'skbuild.constants'
```

Its `setup.py` imports classic internals that the rewrite dropped:

```python
from skbuild.constants import CMAKE_INSTALL_DIR, skbuild_plat_name  # module removed
from skbuild.cmaker import get_cmake_version                        # module removed
```

`skbuild.constants`, `skbuild.cmaker`, and `skbuild.setuptools_wrap` no longer exist on the branch — `__init__` just re-exports `setup` from scikit-build-core. DracoPy uses `CMAKE_INSTALL_DIR()` to build its `include_dirs`/`lib_dirs` and `skbuild_plat_name()` to derive the macOS deployment target/arch, so there is no trivial drop-in. This is deeper than the usual `from skbuild import setup` surface, but these were public modules in classic scikit-build. Worth deciding whether to ship compat shims for `skbuild.constants` / `skbuild.cmaker` or document the break. Note the CMake-side compat vars (`SKBUILD`, `PYTHON_VERSION_STRING`, `PYTHON_EXECUTABLE`, …) *are* still set — pico_tree relies on them and built fine — so this gap is only the Python-level modules.

### Also of note — install scope is narrower (usually an improvement)

Related to the symlink note above: the core backend ships only what `cmake --install` actually emits. For ezc3d the classic wheel bundled ~54 stray source files into `.data/data/` (`src/*.cpp`, `setup.py`, even a `*.h.in` template CMake explicitly EXCLUDEs); the branch wheel omits them — a cleanup. Same mechanism as the python-blosc symlink drop.

### OK, no regression
- **foldcomp** (C++) — identical wheel contents, imports
- **pico_tree** (pybind11) — identical wheels, KdTree works; relies on legacy CMake vars, intact
- **cryptomite** (pybind11) — `.so` byte-identical, imports
- **ezc3d** (SWIG) — identical importable package (see cleanup note above)

All builds macOS arm64, single Python each, `--no-build-isolation`.

henryiii:

**Census: how many known downstreams import removed `skbuild` internals.** Scanned every `setup.py` (any depth) in the 167 skbuild-using repos from a Sourcegraph search (excluding our own repos + BSD ports collections), grepping for dotted `skbuild.` references on each project's current default branch.

**Result: 12 of 167 (~7%) import a module the branch removed.** The branch keeps only `skbuild.setup`, `skbuild.exceptions.SKBuildError`, and `skbuild.__version__`; everything else (`constants`, `cmaker`, `setuptools_wrap`, `command.*`, `utils`, `platform_specifics`) is gone.

Breakers, by what they import:

| Removed module | Repos |
|---|---|
| `skbuild.constants` (`CMAKE_INSTALL_DIR`, `CMAKE_BUILD_DIR`, `skbuild_plat_name`, `set_skbuild_plat_name`) | whisper-cpp-python, bgfx-python, otterbrix, EGTTools, taichi, DracoPy |
| `skbuild.cmaker` (`get_cmake_version`; opencv also `from skbuild import cmaker`) | asgard, theengs/decoder, DracoPy, opencv-python |
| `skbuild.command.*` (`sdist`, `egg_info`, `build_ext`) | dumux, taichi, vaex |
| `skbuild.setuptools_wrap` | dpctl, opencv-python |
| `skbuild.utils` | dpctl |

Notes:
- The other 155: 130 use only the public surface (`from skbuild import setup`) and are fine; 14 matched only kept symbols (`setup`/`exceptions`/`__file__`); 10 have no plain `setup.py` (templated `setup.py.in` or already migrated — e.g. netgen/ngsolve, PDAL, Slycot); 1 (TileDB-Py) was a false match — its `skbuild.cmake.define...` are scikit-build-**core** `-C` config options, already migrated.
- Only bites a project that actually installs scikit-build >=1.0, i.e. after it widens its version pin — not automatic.

**Cheap shims cover most of it:** a `skbuild.constants` shim (a handful of path/plat-name helpers) unblocks 6, and `skbuild.cmaker.get_cmake_version` unblocks 4 (overlap: DracoPy). `command.*` subclasses and `setuptools_wrap` are the hard ones (3 + 2 repos) and less worth reproducing. Reproducers for all 12 available if useful.

henryiii:

Prototyped compat shims for the two most-used removed internals, then trimmed them to only the parts that stay correct under scikit-build-core.

**Kept (pure, layout-independent):**
- `skbuild.constants.skbuild_plat_name()` — computes the platform tag; no dependency on where the build happens.
- `skbuild.cmaker.get_cmake_version()` — runs `cmake --version`, raises `SKBuildError` on failure.

Both emit a `DeprecationWarning` and pass strict mypy + ruff.

**Deliberately dropped:** the `CMAKE_INSTALL_DIR()` / `CMAKE_BUILD_DIR()` path helpers. They described the old `_skbuild/<plat>-<pyver>/` layout, which scikit-build-core does not build into, so shimming them would hand callers a clean import that silently points at a non-existent directory — worse than a clear `ImportError`. Consumers that read files from those paths (otterbrix, EGTTools, DracoPy) need real porting, not a shim.

**What the two shims fix, of the 12 downstreams that import removed internals:**
- Fully: bgfx-python (`skbuild_plat_name`), asgard + theengs/decoder (`get_cmake_version`), opencv-python's `from skbuild import cmaker` line.
- Partially: DracoPy, taichi — their version/plat-name uses work, but they also import `CMAKE_INSTALL_DIR` / `command.egg_info`, which we do not provide.
- Not covered (by design): otterbrix, EGTTools, dpctl, vaex, dumux — these lean on `setuptools_wrap` / `command.*` / the path helpers.

Happy to open this as a draft PR if we want the two shims in 1.0; otherwise it stays a documented option.

henryiii:
Shims added for `skbuild.constants.CMAKE_INSTALL_DIR()` (with caveats), `skbuild.constants.skbuild_plat_name()`, and `skbuild.cmaker.get_cmake_version()`. All currently also show deprecation warnings. That should fix the easy ones above (tested on DracoPy). The command ones are much harder to fix, so those will probably have to be casualties - setuptools deprecated it's command internals a long time ago, so I think it's fair.

henryiii:

Added import stubs for `skbuild.setuptools_wrap` and `skbuild.utils` (1ce1570).

These modules expose **nothing** — a bare `import skbuild.setuptools_wrap` (some projects keep one for historical reasons) succeeds, but every attribute access raises `AttributeError` via a module-level `__getattr__`. The point is to draw a clean line:

- **Bare importers pass.** `IntelPython/dpctl` imports both modules and never dereferences them; it now builds. That drops the downstream census of still-broken internal users from 6 to 5.
- **Internal users fail loudly.** `opencv-python` monkeypatches `skbuild.setuptools_wrap._classify_installed_files` and calls `_copy_file` — the classic install-classification pipeline core doesn't have. Re-exporting those would let the monkeypatch install cleanly and then silently never fire, producing a wrong wheel. Instead the attribute read raises `AttributeError` pointing at scikit-build-core, so it breaks at the exact line it always relied on rather than quietly no-op'ing.

Remaining internal users needing a real port (not a shim): `opencv-python` and `dpctl`'s peers on `setuptools_wrap`, plus `command.*` users (`dumux`, `vaex`, `taichi`) and `constants.CMAKE_BUILD_DIR`/`set_skbuild_plat_name` (`otterbrix`, `taichi`). Those need hook points in core, not compat stubs here.

henryiii:
Of the 12 using internals, 6 pass now with shims, and I've opened issues on the 6 others. I've only tried a subset of the 167 projects outside of those, but without internals access they mostly pass or work better with the swap.

PR #1014, opened 2023-07-21
chore: remove pkg_resources and test command

henryiii:
- fix: drop pkg_resources
- fix: drop custom test command

PR #1195, opened 2026-06-22
feat: deprecate classic features removed by the scikit-build-core backend swap

henryiii:

## Why

The next major release (#1185) replaces the classic backend with scikit-build-core's setuptools plugin, which removes a number of features outright or turns them into errors. This adds deprecation warnings to a **final classic release** so users hit the warnings *before* upgrading breaks them.

All warnings use a new `skbuild.exceptions.SKBuildDeprecationWarning(FutureWarning)`, so users/CI can filter scikit-build's deprecations precisely while keeping them shown-by-default.

Scope was deliberately limited to two categories (CLI options and sdist/editable behavior were left out):

### 1. `setup()` keyword arguments

Warned only when set to a non-default value:

| kwarg | warns when | behavior in #1185 |
|---|---|---|
| `cmake_with_sdist` | `True` | becomes an error |
| `cmake_install_target` | `!= "install"` | becomes an error |
| `cmake_languages` | `!= ("C", "CXX")` | ignored |
| `cmake_minimum_required_version` | not `None` | ignored |

### 2. Internal module imports

`cmaker`, `constants`, `utils`, `command`, `platform_specifics`, and `setuptools_wrap` are removed in #1185. Because `setuptools_wrap` eagerly imports them and `__init__` re-exports `setup`, a naive top-level `warnings.warn` would fire on every normal `import skbuild`. So each real module was moved to a private `_X`, and the public name is now a thin shim (`from ._X import *` + warn). All internal imports and `__init__.py` were repointed to `_X`, so the documented `from skbuild import setup` stays **silent** while `import skbuild.cmaker` (etc.) warns.

## Tests & docs

- `pyproject.toml` globally ignores `SKBuildDeprecationWarning` (the suite imports these internals everywhere, including `tests/__init__.py`).
- New `tests/test_deprecations.py` asserts every warning explicitly — `pytest.warns` for kwargs, a subprocess per module for imports (module caching means an in-process re-import wouldn't re-fire).
- Repointed four test files whose mocks targeted now-private internals.
- Docs autodoc repointed to the private modules; removed a long-dead `skbuild.command.test` autodoc reference (the module was deleted in #1014; its doc stub had been silently breaking the `-W` docs build).

## Verification

- `prek -a` (ruff + strict mypy): green.
- Full suite (`-m "not isolated and not fortran"`): 211 passed. The one failure (`test_issue707::test_install_command`) reproduces identically on `main` — a Python 3.14t free-threading GIL `RuntimeWarning`, unrelated to this change.
- `tests/test_deprecations.py`: 12 passed.
- Confirmed the happy path (`from skbuild import setup`) is warning-free and each deprecated path warns.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


henryiii:
Codex GPT 5.5 review

The patch intends to deprecate internal public module paths before removal, but moving package submodules without shims breaks existing documented imports immediately. This is a backwards-compatibility regression for projects importing command or platform submodules.

Full review comments:

- [P1] Preserve deprecated command submodule imports — /Users/henryfs/git/scikit-build/scikit-build/skbuild/command/__init__.py:12-12
  For projects that import documented command submodules directly, e.g. `from skbuild.command.build_ext import build_ext`, this alias package now only re-exports symbols from `skbuild._command` and leaves no `skbuild.command.build_ext` module on the package path, so the import raises `ModuleNotFoundError` instead of emitting the planned deprecation warning. Please keep shim submodules or otherwise register aliases for the old `skbuild.command.*` module names until the major release that removes them.

- [P1] Preserve deprecated platform submodule imports — /Users/henryfs/git/scikit-build/scikit-build/skbuild/platform_specifics/__init__.py:12-12
  For users importing platform-specific helpers directly, e.g. `from skbuild.platform_specifics.windows import find_visual_studio`, this package alias does not provide the old `skbuild.platform_specifics.*` submodules after the files were moved to `skbuild._platform_specifics`, so those imports fail immediately rather than warning. Please add compatibility shims or module aliases for the old platform submodule names during the deprecation period.

PR #1223, opened 2026-07-24
ci: fetch full history for readthedocs builds

henryiii:

Read the Docs was showing the version as `0.1dev50`. RTD does a shallow clone with no tags, so setuptools-scm falls back to a dev version. Adding a `post_checkout` job to unshallow and fetch tags fixes it — the same fix used in scikit-build-core.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -13,6 +13,11 @@ build:
   os: ubuntu-24.04
   tools:
     python: "3.13"
+  jobs:
+    post_checkout:
+      # RTD does a shallow clone, hiding the release tag from setuptools-scm and
+      # producing a bogus 0.1.devN version. Fetch full history + tags.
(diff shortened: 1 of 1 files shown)


henryiii:
No logs, but these PRs aren't causing the errors, I'm pretty sure. At least this one isn't (the other messes with extras, though I think I've handled that). CC @LecrisUT.

LecrisUT:
Oh, huh, this is a new one. Smells like an outage. And on a Friday evening of course 🙃

LecrisUT:
Seems like it was an incident. Seems to have recovered in the meantime https://copr.fedorainfracloud.org/coprs/g/scikit-build/nightly/build/10771514/

henryiii:
Okay, great! GitHub had an incident earlier today.

PR #1228, opened 2026-07-30
fix(skbuild): guard against missing `_version.py`

### LINKED ISSUE #1227: No module named 'skbuild._version' when building docs in a tagless fork
I tried to build the documentation in a forked repository using the following command:

```bash
uv run --extra docs make -C docs html
```

However, it failed with a `ModuleNotFoundError` stating that `skbuild._version` could not be found:

```python
Traceback
=========

      File "/media/hwhsu1231/KubuntuData/Repo/hwhsu1231-fork/scikit-build/.venv/lib/python3.14/site-packages/sphinx/config.py", line 611, in eval_config_file
        raise ConfigError(msg % traceback.format_exc()) from exc
    sphinx.errors.ConfigError: There is a programmable error in your configuration file:
    
    Traceback (most recent call last):
      File "/media/hwhsu1231/KubuntuData/Repo/hwhsu1231-fork/scikit-build/.venv/lib/python3.14/site-packages/sphinx/config.py", line 596, in eval_config_file
        exec(code, namespace)  # NoQA: S102
        ~~~~^^^^^^^^^^^^^^^^^
      File "/media/hwhsu1231/KubuntuData/Repo/hwhsu1231-fork/scikit-build/docs/conf.py", line 9, in <module>
        import skbuild
      File "/media/hwhsu1231/KubuntuData/Repo/hwhsu1231-fork/scikit-build/skbuild/__init__.py", line 12, in <module>
        from ._version import version as __version__
    ModuleNotFoundError: No module named 'skbuild._version'
```

The root cause of this error is that the forked repository does not contain Git tags, which causes `hatch` to skip generating the `skbuild/_version.py` file during dependency installation:

https://github.com/scikit-build/scikit-build/blob/0cbfb0eeea06a697c3e40d8c561c5639ae0aa22f/pyproject.toml#L76

Although fetching tags from the upstream repository resolves the issue, it is quite inconvenient. Therefore, I suggest updating `skbuild/__init__.py` from:

https://github.com/scikit-build/scikit-build/blob/0cbfb0eeea06a697c3e40d8c561c5639ae0aa22f/skbuild/__init__.py#L12

to:

```python
try:
    from ._version import version as __version__
except ImportError:
    __version__ = "0.0.0.dev0"
```

This way, even if `skbuild/_version.py` is missing when building the documentation, it can fall back to a default version number like `0.0.0.dev0` as a workaround.

hwhsu1231:
When `hatch-vcs` cannot determine a version (e.g., in fork repositories without git tags), `_version.py` is not generated, causing a `ModuleNotFoundError` at import time. Add a `try`/`except` fallback so the package remains importable for development and documentation builds.

Close: #1227

### DIFF
--- a/skbuild/__init__.py
+++ b/skbuild/__init__.py
@@ -9,7 +9,10 @@ from __future__ import annotations
 
 from scikit_build_core.setuptools.wrapper import setup
 
-from ._version import version as __version__
+try:
+    from ._version import version as __version__
+except ImportError:  # pragma: no cover
+    __version__ = "0.0.0.dev0"
 
 __author__ = "The scikit-build team"
 __email__ = "scikit-build@googlegroups.com"


henryiii:
You can set a version via environment variable (SETUPTOOLS_SCM_PRETEND_VERISON or something like that), which will print out nicer in the docs too. Does that work instead?

henryiii:
(I'm not against this, to be clear, though the coverage complaint is an issue - it just might be nicer on your end to set the variable for the docs also.)

Edit: I thought this was in docs. Little worse being in source, it might hide issues where a packager missed producing this but we don't want to report version 0.

hwhsu1231:
Closing this PR as further testing revealed that the root cause of #1227 is actually completely unrelated to tags.

